PR #2, opened 2016-07-13
Add word clock

TheLastProject:
Also fixes up the analog clock a bit (should improve performance)

002-words.qml is based on 000-default-digital.qml and therefore uses the same copyright header (with my name added)


TheLastProject:
I am observing horrible battery drain. Not sure if it is this watchface. May need more research.

PR #10, opened 2017-02-19
Add activity watchface

aebruno:
The idea for this watchface is to display the value of step counter
against a daily goal using the pedometer sensor in bass. Until sensors
are fully supported in AsteroidOS we display the current battery charge
level as a proof of concept.


FlorentRevest:
Hi, 
First of all this is a good idea. However I won't merge it for several reasons:
- there is already a circular progress component in qml-asteroid, you should use/improve it instead of recreating your own https://git.io/vD5Lt
- without shadows your text is impossible to read on white backgrounds. 
- this doesn't support sensors yet. (which would be more useful)
- you should consider working with a project's designer on how to choose a color different than red or this kind of thing. I can put you in touch with one of them to discuss the idea. 
Anyway, good job, i can't wait to see a final version of that merged !

aebruno:
All sounds good! Thanks for the great feedback. I'll try and get some of these improvements implemented.

FlorentRevest:
Hi aebruno. I like the new watchface much more good job! Please refer to the second paragraph of this post https://github.com/AsteroidOS/asteroid-launcher/pull/11#issuecomment-282874825 which is also completely relevant to your pull request.

PR #11, opened 2017-02-27
add 006-classic-analog watchface

anyc:
Based on 001-analog, I added small hours and minute marks and a seconds hand. For the additional hand, I changed the update frequency to "seconds". I do not know, however, if this has any negative implications.

It is not my favorite watchface but I think it would serve well as an example for a more classic look.

By the way, I do not know if this was intentional, but in this watchface I removed the ctx.closePath() calls as they made the hand longer if I just increased the lineWidth. This was a bit suprising to me. Without, the lineWidth and the hand length act more predictable.


anyc:
If I have some time again, I will also fix the wrong drop shadow. ;)

FlorentRevest:
Good job anyc. I like your work but I have two requests:
- On the technical side I would want to know if having used Canvas makes you notice any significant loss of battery life. Canvas is usually very bad for heavy operations because it reinterprets javascript everytime.
- On the "artistic" side I find it very nice but I feel like I'm not the best person to decide which watchfaces should or shouldn't enter the default set of asteroid-launcher. In order to keep track of community watchfaces anyway I decided to create another repository at https://github.com/AsteroidOS/unofficial-watchfaces Resubmit your PR there and I will accept it. If the designers find your work suitable with little to no modification it can eventually enter "upstream" otherwise it can also give ideas to other developers.
Thanks

PR #12, opened 2017-03-22
Fix for watchface not updating

KorakSilvercloud:
Adding in the call to repaint the watchface when time changes.


FlorentRevest:
Thanks!

PR #15, opened 2017-04-16
Handle notification icons that come from syncing

daniellandau:
On my setup the icons from e.g. usb cable connection events seem to come in `notification.icon`, but the ones from the Android sync clients come in `notification.appIcon`. This code handles both.


FlorentRevest:
Good job, I noticed the incorrect behavior of android's notification but haven't had the time to find a solution :) thanks for submitting upstream

PR #21, opened 2017-05-25
fix dropshadows in analog watchface

anyc:
Instead of rotating the hands and their dropshadow from 12 o'clock position, this patch causes the hands to be drawn directly with their right angle.


anyc:
Ah, I almost forgot, I just tested this with qmlscene as I can't use my watch right now.

FlorentRevest:
I have been able to test it on a watch and it works great. Thanks!

Issue #25, opened 2017-11-10
Some watchfaces show 00 instead of 12 when "Use 12H format" is on


skylarmt:
According to [this Wikipedia article](https://en.wikipedia.org/wiki/12-hour_clock) and living in the USA, clocks go from 11:59pm to 12:00am at midnight, not 11:59pm to 00:00am.

This line of code is the culprit:  
`if(use12H.value) hour = hour % 12`  
If `hour` is `0`, `12`, or `24`, it will be set to `0`.  Both `/watchfaces/000-default-digital.qml` and `/watchfaces/004-alternative-digital-2.qml` are affected, as they both use this code.

The other digital watchfaces are not affected, as they use [Qt's formatDateTime() function](http://doc.qt.io/qt-4.8/qml-qt.html#formatDateTime-method), which automatically formats hours as "00 to 23 or 01 to 12 if AM/PM display".

The Time setting UI is also affected, as it shows `00` instead of `12`.

FlorentRevest:
Thank you for the notice. It's added on my TODO List, I'll let you know once it will be solved.

skylarmt:
I'm working on setting up a dev environment so I can fix it, but my Internet is incredibly slow, so you might get to it first.

PR #26, opened 2017-11-10
Fix hour 12 displaying as "00" in 12-hour unit mode on some watchfaces.

Huntereb:
#25


FlorentRevest:
Hi Huntereb,
Please fix indentation.
Also, consider using http://doc.qt.io/qt-4.8/qml-qt.html#formatDateTime-method when possible as it's much cleaner that what we have here.

Huntereb:
This what you had in mind?

PR #41, opened 2020-07-02
watchfaces: Fix hour hand not moving when minute changes.

MagneFire:
The hour hand needs to move slightly when the minute hand changes.
This only worked when setting a new watchface(i.e. onCompleted), not when the time actually changes.

Please let me know if I am the only one experiencing this issue. As this is something that is quite noticeable.

### DIFF
--- a/watchfaces/002-analog-70s-classic.qml
+++ b/watchfaces/002-analog-70s-classic.qml
@@ -329,11 +329,11 @@ Item {
             if(secondHand.second != second) {
                 secondHand.second = second
                 secondHand.requestPaint()
+            }if(hourHand.hour != hour) {
+                hourHand.hour = hour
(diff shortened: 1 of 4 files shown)


FlorentRevest:
Oh good catch! :) I guess it could have been "by design" to have the clock handle move once per hour but still, what you propose makes more sense. Thanks!

eLtMosen:
Thx @MagneFire !
That was indeed not what i intended and a clear oversight on my side.
Thx again for fixing directly.

MagneFire:
@eLtMosen Thanks for acknowledging this :)
I thought that this was intended, because I heard no one crying about this issue :wink: 
Also thanks for the awesome watchfaces :+1:

PR #45, opened 2020-09-20
MainScreen: Show notification indicator in ambient mode.

MagneFire:
Currently it is not possible to see if you have new notifications in aod mode.
~This patch enables the left indicator in aod mode only when a new notification is added. It will be hidden when the notifications have been read.
Also adds a simple burn in prevention method: move the indicator left-right every time the screen gets updated in aod mode.~

Preview:
Screenshot_20200920_203449

~Edit: Don't merge! After testing I noticed that this does not take into account when a notification has been removed from the sync client! ( I will probably adjust the notification list model in lipstick to keep track of not viewed notifications.)~

As per the discussion I've changed the implementation such that it move the left indicator in all directions. It will now also be always visible when there are notifications. @daniellandau Thanks for your feedback!

One thing worth a note is that all indicators will animate when the screen awakes. The purpose is two fold: Makes the screen less static when waking and move back the left indicator in the horizontal direction.

Here is another preview with the burn-in-protection in action:
ezgif-5-c249ff5e826c

### DIFF
--- a/qml/MainScreen.qml
+++ b/qml/MainScreen.qml
@@ -151,15 +151,24 @@ Item {
             wallClock.enabled = true
         }
         onDisplayAboutToBeOff: wallClock.enabled = false
-        onDisplayOn: if (Lipstick.compositor.ambientEnabled) grid.moveTo(0, 0)
+        onDisplayOn: {
+            grid.animateIndicators()
+            if (Lipstick.compositor.ambientEnabled) grid.moveTo(0, 0)
+        }
         onDisplayAmbientChanged: wallpaperAnimation.duration = 300
-        onDisplayAmbientEntered: wallpaperDarkener.opacity = 1;
-        onDisplayAmbientLeft: wallpaperDarkener.opacity = 0;
+        onDisplayAmbientEntered: wallpaperDarkener.opacity = 1
+        onDisplayAmbientLeft: {
+            wallpaperDarkener.opacity = 0
+            if (burnInProtectionManager.enabled) leftIndicator.anchors.verticalCenterOffset = 0
+        }
         onDisplayAmbientUpdate: {
             // Perform burn in protection
             if (burnInProtectionManager.enabled) {
                 grid.contentX = Math.random()*(burnInProtectionManager.leftOffset + burnInProtectionManager.rightOffset)-burnInProtectionManager.leftOffset
                 grid.contentY = Math.random()*(burnInProtectionManager.topOffset + burnInProtectionManager.bottomOffset)-burnInProtectionManager.topOffset
+
+                leftIndicator.anchors.horizontalCenterOffset = (Math.random()*2)*leftIndicator.finWidth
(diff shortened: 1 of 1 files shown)


FlorentRevest:
Not merging as per the description.

I don't have specific comments on the changes. This just leaves me with a feeling that asteroid-launcher is written with iterative reflexes in mind and does not take full benefits from the declarative paradigm. I don't have anything actionable to propose though.

qml/MainScreen.qml
```diff
@@ -199,7 +204,7 @@ Item {
             addPanel(0, -1, topPanel)
 
             rightIndicator.visible  = Qt.binding(function() { return ((grid.toLeftAllowed   || (grid.currentVerticalPos == 1 && al.toLeftAllowed )) && !displayAmbient)})
-            leftIndicator.visible   = Qt.binding(function() { return ((grid.toRightAllowed  || (grid.currentVerticalPos == 1 && al.toRightAllowed)) && !displayAmbient)})
+            leftIndicator.visible   = Qt.binding(function() { return ((grid.toRightAllowed  || (grid.currentVerticalPos == 1 && al.toRightAllowed)) && (!displayAmbient || np.hasNew))})
```
daniellandau, line 207: What's the point of keeping track of new/seen notifications? Wouldn't it be simpler to just have `!np.modelEmpty`?

daniellandau: I tested it now like that, and it removes the indicator when notification is removed on phones side. You can then leave `NotificationsPanel.qml` unchanged.

MagneFire: Thanks for your feedback! :smile: 
What you mean will indeed work. However with `!np.modelEmpty` you do not encompass when a user views a notification, as in the user swipes left, then does not dismiss it. This will make the indicator stay visible, while the user has seen the notification already.
What I want to achieve is a behavior that when a user has seen the notification but not dismissed it. There is no need to keep showing the indicator when the user has seen it but not dismissed it. This should improve the screen with respect to burn in problems.

daniellandau:
I understand, but disagree on the importance of that logic. I don't ever go to the notifications area and leave stuff undismissed, and if I do it's because I want to be maximally reminded of them.

I think we should worry about burn-in as far as it doesn't interfere with UX, but not more.

MagneFire:
Hmm, ok. The pros of using your suggested solution are as follows:
- Easy implementation.
- Less code fragmentation (Lipstick needs changes for my proposed method).
- Consistent with the current behavior.

So yeah, the only negative thing I can think of is the burn in problem.

Here is my suggestion for a suitable solution:
- Use your proposed method.
- Add vertical burn-in protection as well.

That should, hopefully, make it safer for the screen for people (unlike you :wink: ) that keep the notifications without dismissing them.

daniellandau:
Sounds good to me 👍🏼

daniellandau:
LGTM

PR #47, opened 2021-02-19
GestureFilterArea: Don't prioritize horizontal swipes.

MagneFire:
For some reason this issue is more noticeable depending on the watch used. For example, swiping in any direction works great on both `sturgeon` and `tetra`, but on others a horizontal swipe occurs to be preferred even though a vertical swipe is intended, this is something that I notices on `smelt` and to a lesser degree on `swift`. There are also reports that this happens on the wip port for `sawfish` (https://github.com/AsteroidOS/asteroid/issues/101#issuecomment-778785938 and https://github.com/AsteroidOS/asteroid/issues/101#issuecomment-778827054).

Tested on `smelt` as it was really bad there. With this patch it seems to work perfectly (i.e. the same as `sturgeon` now :smile: )


FlorentRevest:
Great! I remember it also happened on harmony

PR #59, opened 2021-06-03
fixed 12H setting behaviour of 4 stock watchfaces

eLtMosen:
In effort to fix all watchfaces 12H behaviour as reported here:
https://github.com/AsteroidOS/unofficial-watchfaces/issues/19

### DIFF
--- a/watchfaces/002-analog-70s-classic.qml
+++ b/watchfaces/002-analog-70s-classic.qml
@@ -58,7 +58,10 @@ Item {
             top: dayDisplay.bottom
             topMargin: parent.height * 0.0156
         }
-        text: wallClock.time.toLocaleString(Qt.locale(), "<b>HH</b>:mm")
+        text: if (use12H.value) {
+                  wallClock.time.toLocaleString(Qt.locale(), "hh ap").slice(0, 2) + wallClock.time.toLocaleString(Qt.locale(), ":mm") }
+              else
+                  wallClock.time.toLocaleString(Qt.locale(), "HH:mm")
     }
 
     Text {
@@ -327,12 +330,12 @@ Item {
             var hour = wallClock.time.getHours()
             var minute = wallClock.time.getMinutes()
             var second = wallClock.time.getSeconds()
-            if(secondHand.second != second) {
+            if(secondHand.second !== second) {
                 secondHand.second = second
                 secondHand.requestPaint()
-            }if(hourHand.hour != hour) {
+            }if(hourHand.hour !== hour) {
                 hourHand.hour = hour
-            }if(minuteHand.minute != minute) {
+            }if(minuteHand.minute !== minute) {
                 minuteHand.minute = minute
                 minuteHand.requestPaint()
                 hourHand.requestPaint()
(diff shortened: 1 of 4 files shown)


watchfaces/002-analog-70s-classic.qml
```diff
@@ -58,7 +58,10 @@ Item {
             top: dayDisplay.bottom
             topMargin: parent.height * 0.0156
         }
-        text: wallClock.time.toLocaleString(Qt.locale(), "<b>HH</b>:mm")
+        text: if (use12H.value) {
+                  wallClock.time.toLocaleString(Qt.locale(), "hh ap").slice(0, 2) + wallClock.time.toLocaleString(Qt.locale(), ":mm") }
+              else
+                  wallClock.time.toLocaleString(Qt.locale(), "HH:mm")
```
FlorentRevest, line 64: it looks like you dropped the <b> </b> ?

watchfaces/007-bold-hour-bebas.qml
```diff
@@ -66,7 +66,10 @@ Item {
         horizontalAlignment: Text.AlignHCenter
         anchors.horizontalCenter: parent.horizontalCenter
         anchors.verticalCenter: parent.verticalCenter
-        text: wallClock.time.toLocaleString(Qt.locale(), "HH")
+        text: if (use12H.value) {
```
FlorentRevest, line 69: Mh, this is a lot of duplicated code :/ What about exposing a function in asteroid-launcher along the line of:

```
function paddedHour(time) {
  if (use12H.value)
    return time.toLocaleString(Qt.locale(), "hh ap").slice(0, 2)
  else
    return time.toLocaleString(Qt.locale(), "HH")
}
```

And then use it as `paddedHour(wallClock.time)` when only the time is shown or in more complex cases something like: `"<b>" + paddedHour(wallClock.time) + "</b>"  + wallClock.time.toLocaleString(Qt.locale(), ":mm"`

I think this would make the code more readable and a bit easier to maintain

eLtMosen: Good Idea!
I was about to have a function in every watchface but discarded that idea since it would even add more code lines for the function syntax.
How to go about? Will you implement the function so i can rework the watchfaces and update the PR?

watchfaces/002-analog-70s-classic.qml
eLtMosen: If you mean i dropped the bold tags. Yes, those where having no use since the League spartan font can not be made heavier that way. It could be manipulated with the font.weight: value but as it is now, there should be no change visually.

FlorentRevest: ah alright :) if there is no visual change then it's all good

watchfaces/007-bold-hour-bebas.qml
FlorentRevest: Ah that's a good question indeed, let me build you a modified asteroid-launcher and if you can get it to work then I push the change and then you can push your PRs :)

FlorentRevest: so I started to implement the function in asteroid-launcher but realized it feels really out of place there so I'll just merge the PR ;) sorry for the delay and thank you for the contrib!

PR #60, opened 2021-06-03
More lag fixes

MagneFire:
Basically I observed too more issues with the previous work: panels would hide while animation was still active and the UI would lag a lot when many notifications arrive at once.

Important! This now adds another loop over all panels. In my opinion this makes the code a bit nasty...

If you have another method or different method to implement the panel grid altogether, then please let me know too :wink:


FlorentRevest:
Ugh, yes, the code is extremely nasty, it would benefit from a complete rewrite at some point, but given that it's already a pile of hacks, it's ok to stack some more on top of it :p

PR #62, opened 2021-06-13
fixed .requestPaint for the numberStrokes item in Ambient mode

eLtMosen:
small correction so the prepared ambient colour change in numberStroke item is actually executed.


FlorentRevest:
Good catch! :)

Issue #61, opened 2021-06-13
[Discussion] Idea for a new "app-drawer" more suitable for a watch


eLtMosen:
I don't find the app drawer concept suitable for a watch.
Its operation requires constant eye contact. Is imo not precise in reaching a certain app. Favourite app positions are hard to memorize in a longer list. In the worst case needs repeated finger re-placement on the screen to scroll through.
We have lost a feature on smartwatches we where used to have back in the Casio days.
Blind operation aided by haptic feedback.
Though we can not bring back hardware buttons, maybe we can give our brain and fingers a new easy to memorize pattern that allows for some blind operation on touch-screens of defined size.
Like, starting up an app even before making full eye contact with the watch while bringing it into readable distance.

#pseudo algorithmic:  

1) User does a swipe up from the bottom of the screen to the full top 12 o’clock position and keeps holding the finger on the screen.

2) Instead of bringing up the stock app-drawer, only the watchface fades out during that swipe up across the screen and makes place for the first app starter icon to fade in below the finger.

3) From this 12 o’clock position, still holding the finger on screen, user continues to do a circular swipe along the outer edge of the (ideally round) screen. Either in left or right direction according to left- right handed.

4) With every ¼ rotation of the swipe around the screen, the next app „locks“ into start position. A short vibration of the watch could confirm every ¼ rotation additionally.

5) The corresponding app gets started as soon as the user lifts up the finger from the screen.

6) The order of the apps should be configurable, so that users can sort their favourites in the first 4 quarter turns since those are best memorable.

7) As hinted by @jrtberlin, the ¼ rotation lock paradigm to select apps could easily be used on „crowns“ some smartwatch models are equipped with.

## Pros
- Reduce the overall complexity of the UI
- Have an app-drawer system that is unique to the use-case and closer to a traditional watch experience
- App selection takes place on one screen, giving the impression the apps being core features rather then 3rd party add on
- One continuous finger movement in defined environment removes necessity for eye contact.
- Finger positions to start an app are fixed and thus easy to memorize.
- ¼ turns around a circle are easy to haptically memorize
- haptic feedback from vibration helps with memorizing
- 4 favourite apps are sufficient for most users, further apps are still accessible with more ¼ turns.
- „Finger hiding the screen content“ issue is removed for the memorized app-starters

## Cons
[...]

PR #63, opened 2021-08-08
Add support for dynamic app launchers.

MagneFire:
This PR adds support for different app launcher styles. Eventually, it might even help to implement this awesome idea: https://github.com/AsteroidOS/asteroid-launcher/issues/61

Currently, I'm looking for feedback on this code as there is quite some code duplication going on between the different app launcher styles. As well as there is a need for explicit definitions for the `to*Allowed` and the `forbit*` variables.

At the moment this work also appears to have broken the tutorial of the launcher. I will need to investigate this too.

@LittleFox94 I've also added your app launcher style in a separate commit (Of which you are the author). Please let me know if you are ok with this, otherwise I'll adjust it to what you want or remove it entirely.

Demo:

https://user-images.githubusercontent.com/7857908/128628281-fd9963be-e07f-44c0-b5b9-bc6f3020d020.mp4

This finally implements https://github.com/AsteroidOS/asteroid/issues/152.

Current TODOs:
- ~The tutorial is broken as `contentX` doesn't exists on the Loader class, need to find a better solution.~ Solved by means of setting the `Desktop.appLauncher` once loaded: https://github.com/AsteroidOS/asteroid-launcher/pull/63/files#diff-142fd70833d50de90a6f6704a050370378371f9c6ed2fa6716203d1299c5cfdaR227
- ~I actually think that it might be better to nuke the `diskBackground.svg`s and basically revert this commit: https://github.com/AsteroidOS/asteroid-launcher/commit/15c3e53bac6dbe343df04aa378d13de2d423c8b6. This will worsen support on `tetra`, but it has broken watchfaces anyway (The masked spartan for example). And as discussed with @FlorentRevest before, it is not a good design approach to have a worse code base because one watch is misbehaving.~ Also solved by means of: https://github.com/AsteroidOS/asteroid-launcher/pull/63/commits/2490e8447126d9dd6cd224a7faeed836132de763


LittleFox94:
Am fine with my style added here :)

Would like to review, let's see if I can manage to do that .. pls don't block on my feedback though

applauncher-img/diskBackground.svg
```diff
@@ -0,0 +1,14 @@
+<svg width="100" height="100" viewBox="0 0 100 100"
```
FlorentRevest, line 1: Should you delete the ones in images/ ?

qml/MainScreen.qml
```diff
@@ -182,20 +186,43 @@ Item {
...
+        defaultValue: "file:///usr/share/asteroid-launcher/applauncher/000-default-horizontal.qml"
+    }
+
     Connections {
         target: localeManager
         function onChangesObserverChanged() {
             var bkp = watchFaceSource.value
             watchFaceSource.value = ""
             watchFaceSource.value = bkp
+            var bkp1 = appLauncherSource.value
```
FlorentRevest, line 201: let's rename these watchFaceSourceBackup and appLauncherSourceBackup :p

FlorentRevest:
Ugh yeah I see how this is a lot of duplicated code but I can't think of a way around it :| We need to delegate the indicator statuses to the applauncher module and they all behave in a slightly different way so it makes sense to have all that duplicated code...  It's not very elegant but I don't have a better solution.

FlorentRevest:
And just a general comment on the look of the new watchfaces. I am definitely not a designer but I think they would benefit from a bit more padding between the icon and the disk. :) Maybe not as dramatic as the default watchface but still some margin would make it look nicer I think.

applauncher-img/diskBackground.svg
MagneFire: I think that for now it makes sense to keep this indeed here too as it is also used by the `BluetoothAgent` too.

MagneFire:
> And just a general comment on the look of the new watchfaces. I am definitely not a designer but I think they would benefit from a bit more padding between the icon and the disk. :) Maybe not as dramatic as the default watchface but still some margin would make it look nicer I think.

With more padding do you mean increase the size of the disk while keeping the icon the same size or reducing the size of the icon while keeping the disk size identical?

FlorentRevest:
> With more padding do you mean increase the size of the disk while keeping the icon the same size or reducing the size of the icon while keeping the disk size identical?

Both are probably fine :)

applauncher-img/diskBackground.svg
FlorentRevest: Okay. Anyway I agree with "I actually think that it might be better to nuke the diskBackground.svgs and basically revert this commit: 15c3e53. This will worsen support on tetra, but it has broken watchfaces anyway (The masked spartan for example). And as discussed with @FlorentRevest before, it is not a good design approach to have a worse code base because one watch is misbehaving."

applauncher/001-three-icons-horizontal.qml
```diff
@@ -0,0 +1,163 @@
...
+            anchors.fill: parent
+            source: launcherItem.pressed | fakePressed ? "../applauncher-img/diskBackgroundPressed.svg" : "../applauncher-img/diskBackground.svg"
+            sourceSize.width: width
+            sourceSize.height: height
+        }
+
+        Icon {
+            id: icon
+            anchors.centerIn: parent
+            anchors.verticalCenterOffset: -Dims.h(1)
```
FlorentRevest, line 87: I wonder why we keep this offset ? It makes sense in the default style where the text is included below the icon in the disk but here since the icon is the only thing in the disk, it would maybe look better to center the icon ?

applauncher/002-two-columns.qml
```diff
@@ -0,0 +1,176 @@
...
+            source: launcherItem.pressed | fakePressed ? "../applauncher-img/diskBackgroundPressed.svg" : "../applauncher-img/diskBackground.svg"
+            sourceSize.width: width
+            sourceSize.height: height
+            y: -Dims.h(1)
+        }
+
+        Icon {
+            id: icon
+            anchors.centerIn: parent
+            anchors.verticalCenterOffset: -Dims.h(1)
```
FlorentRevest, line 104: Same comment here

MagneFire:
All remaining comments and issues are now solved :smile:

With regards to padding between the icon and the disk, I've slightly increased the size of the disk, here are the before and after comparisons:


Three icon app style:
| Before | After |
|----------|---------|
| Screenshot_20211116_213326 | Screenshot_20211116_213058 |

Two columns app style:
| Before | After |
|----------|---------|
| Screenshot_20211116_213305 | Screenshot_20211116_213120 |

applauncher/001-three-icons-horizontal.qml
MagneFire: I don't think there is a particular reason for this vertical offset. Probably some artifact from the default app launcher style.

applauncher/002-two-columns.qml
MagneFire: Also removed in another revision.

eLtMosen:
> With regards to padding between the icon and the disk, I've slightly increased the size of the disk, here are the before and after comparisons:

Margins look nicely balanced now. Well done!

FlorentRevest:
Thanks, style looks much better indeed

PR #64, opened 2021-08-14
Fixed wrong german syntax in the words watchface

eLtMosen:
As per hint by Allen on the AsteroidOS Matrix server.
Thanx again for spotting and reporting this!


eLtMosen:
Messed a rebase...

PR #68, opened 2021-10-31
Fix invalid French translation in watchface

Oliv6176:
This should fix the invalid French translation AsteroidOS/asteroid#171

My French is quite limited so @TheOneWithTheBraid should probably test if all translations are correct.


TheOneWithTheBraid:
It should be about

- 06:00: `six heures pille`
- 06:05: `six heures ciq`
- 06:10: `six heures dix`
- 06:15: `six heures et quart`
- 06:20: `six heures vingt`
- 06:25: `six heures vingt-ciq`
- 06:30: `six heures et demit`
- 06:35: `six heures trente-cinq` / `vingt-cinq à sept` / `sept heures moins vingt-cinque` - here, depending on the region and person, all three forms are known
- 06:40: `vingt à sept``sept heures moins vingt`
- 06:45: `sept heures moiuns le quart`
- 06:50: `dix à sept` / `sept heures moins dix`
- 06:55: `cinq à sept`

Oliv6176:
@TheOneWithTheBraid Thanks! I have corrected my mistakes.
Is _sept heures moins cinq_ correct for 06:55 as _sept heures moins dix_ is for 06:50?

TheOneWithTheBraid:
Yes, but by far more uncommon than `cinq à sept`. In general, the closer the minute comes to the destination hour, the less probably one would use `DST (heures) moiuns MIN` and the more commonly used is `MIN à DST`.

TheOneWithTheBraid:
Disclaimer: I don't speak metropolitan French.

Oliv6176:
Disclaimer: I don't speak French and can only the bit of French I had at school 😅

Oliv6176:
But for now, should I change it to _cinq à DST_ or keep _DST heures moins cinq_?

TheOneWithTheBraid:
Interesting. Well, I think anyone would understand both. For simplicity, just keep it as it is for now, I'd propose. An option to configure this could be added later - in case anyone is really disturbed.

We can be happy not to handle numbers > 69 for watch faces btw. otherwise we'd definitely need a corresponding setting.

But in general, maybe other applications are interested in whether they should display strings such as `fin-semaine` or `week-end`, `nonante` or `quatre-vingt dix`. As a future prospect, there should be definitely an system-level option to configure language accents.

Same applies for many other languages too: `color` U.S. or `colour` GB, `Januar` DE or `Jänner` AT.

FlorentRevest:
Looks great now, thanks a lot!

PR #69, opened 2021-11-03
MainScreen: Disable burn in protection based on a setting.

MagneFire:
This PR adds support for disabling burn in protection via DeviceInfo a default is setup (i.e. add `MACHINE_PREFER_BURN_IN_PROTECTION = false` to https://github.com/AsteroidOS/meta-dory-hybris/blob/master/conf/machine/dory.conf). Then using Nemos Configuration it can actually be switched on or off.

An additional change could be added to the settings app such that the setting is only visible when `MACHINE_PREFER_BURN_IN_PROTECTION = true`. As this option will always be disabled ONLY for LCDs.

### DIFF
--- a/qml/MainScreen.qml
+++ b/qml/MainScreen.qml
@@ -33,6 +33,7 @@ import Nemo.Time 1.0
 import Nemo.Configuration 1.0
 import org.nemomobile.lipstick 0.1
 import org.asteroid.controls 1.0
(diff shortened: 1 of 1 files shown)


MagneFire:
The additional idea/change has now been applied, when a device doesn't need burn in protection techniques then the option isn't shown at all and disabled in asteroid-launcher.

PR #71, opened 2021-12-08
Improve legibility, Fix spacing issues in settings page

eLtMosen:
To improve legibility,

- `font.styleName` set to SemiCondensed Bold in app launchers 001- and 002-
- DropShadow added to the `icon.Text` in app launchers 001- and 002-
grafik

- During `launcherItem.pressed | fakePressed` not only the circleWrappers color is changed to highlight an input. In 001- and 002- additionally the `icon` gets proportionally darkened to "#444444" as well as the icon.Text in 000-default due to being same color then the icon.
- Centered 002-two-column Items with equal spacing left, right and in between the 2 columns
appdrawer-000-002fix

Solve spacing issues in Launcher settings page

- Change size definition of icontopMargin from `Dim.h(x)` to `parent.height * x` to adapt to smaller environment when shown in Launcher Settings page.
- Same for the `icon` in 000-default.
grafik

### DIFF
--- a/applauncher/000-default-horizontal.qml
+++ b/applauncher/000-default-horizontal.qml
@@ -1,5 +1,6 @@
 /*
- * Copyright (C) 2015 Florent Revest <revestflo@gmail.com>
+ * Copyright (C) 2021 Timo Könnecke <github.com/eLtMosen>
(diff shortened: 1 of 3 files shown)


FlorentRevest:
Very cool stuff! Thanks :D

PR #72, opened 2021-12-17
Fix 008-funky-town-words watchface

eLtMosen:
Correct ambientMode behaviour and better center the minutes in the svg hours.


watchfaces/008-funky-town-words.qml
```diff
@@ -62,19 +62,23 @@ Item {
...
         anchors {
             bottom: parent.bottom
-            bottomMargin: parent.height*0.155
+            bottomMargin: parent.height * 0.15
             horizontalCenter: parent.horizontalCenter
-            horizontalCenterOffset: parent.width*0.210
+            horizontalCenterOffset: parent.width*0.2
         }
-        Behavior on text {
+        Behavior on animToggle {
```
MagneFire, line 81: I think I understand the `displayAmbient` issue: The watchface is trying to animate while the display isn't active hence resulting in a `minuteDisplay` that has a low opacity.

There might be a simpler solution to this though: Why not disable the `Behavior` animation when in `displayAmbient` mode. i.e add a line below: `enabled: !displayAmbient`.

MagneFire (commented):
Liking these fine tuning touches 👌
Only one minor comment from my side, looks great otherwise.

watchfaces/008-funky-town-words.qml
eLtMosen: I tried `running: !displayAmbient`... And then came up withthe workaround when it did not work...
Thanks for the hint.

PR #73, opened 2021-12-19
AppLauncher: Add list app style.

MagneFire:
This adds a new app launcher style.
Thanks to @R0NAM1 and @eLtMosen for the feedback on this one. It improved the look considerable in my opinion.

There are two more changes compared to the one shown on matrix:
- Simplify some of the logic for the text font size (Might be even more poppy).
- Change the opacity of every item such that it disappears when at the top or bottom border.

https://user-images.githubusercontent.com/7857908/146685617-f136c7a1-73a0-480d-b156-4454c424394b.mp4

### DIFF
--- a/applauncher/003-turbo-list.qml
+++ b/applauncher/003-turbo-list.qml
@@ -1,0 +1,194 @@
+/*
+ * Copyright (C) 2021 Darrel Griët <dgriet@gmail.com>
+ *               2021 Timo Könnecke <github.com/eLtMosen>
+ *               2015 Florent Revest <revestflo@gmail.com>
+ *               2014 Aleksi Suomalainen <suomalainen.aleksi@gmail.com>
+ *               2012 Timur Kristóf <venemo@fedoraproject.org>
+ *               2011 Tom Swindell <t.swindell@rubyx.co.uk>
+ * All rights reserved.
+ *
+ * You may use this file under the terms of BSD license as follows:
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of the author nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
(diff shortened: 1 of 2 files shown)


MagneFire:
Updated, renamed the launcher to `turbo-launcher` as semi-suggested by @eLtMosen as well as the incredibly idea to use regular case for this one.

https://user-images.githubusercontent.com/7857908/146833032-b8bd0be1-08b5-467d-8828-7b8debb22eba.mp4

applauncher/003-turbo-list.qml
```diff
@@ -0,0 +1,205 @@
...
+            width: parent.width
+            anchors.leftMargin: parent.width * 0.04
+            anchors.verticalCenter: parent.verticalCenter
+            font.pixelSize: (Math.exp((normalizedOffset)) - 1) * viewScale * Dims.l(6)
+            font.letterSpacing: Dims.l(0.2)
+            font.styleName: "Bold"
+            style: (normalizedOffset >= 0.99) ? Text.Outline : Text.Normal
+            styleColor: alb.centerColor(launcherModel.get(index).filePath)
+            text: model.object.title + localeManager.changesObserver
+        }
```
FlorentRevest, line 181: indentation is a bit borked

applauncher/003-turbo-list.qml
```diff
@@ -0,0 +1,205 @@
...
+        enabled: !appsView.dragging
+        opacity: normalizedOffset
+
+        onClicked: model.object.launchApplication()
+
+        Item {
+            width: parent.width
+            height: parent.height
+            anchors.left: parent.left
+            anchors.leftMargin: DeviceInfo.hasRoundScreen ? offset + Dims.w(5) : 0
```
FlorentRevest, line 118: nice :)

applauncher/003-turbo-list.qml
```diff
@@ -0,0 +1,205 @@
...
+
+    Connections {
+        target: grid
+        function onNormalizedVerOffsetChanged() {
+            // Override wallpaper darkening, so that the default flatmesh color is used.
+            if (!displayAmbient) {
+                wallpaperAnimation.duration = 0
+
+                wallpaper.anchors.verticalCenterOffset = height*grid.normalizedVerOffset*(-0.05)
+                wallpaperDarkener.opacity = Math.abs(grid.normalizedVerOffset)*0.4
```
FlorentRevest, line 66: Ugh, this is very hacky. I would rather have the main asteroid-launcher codebase offer an easy boolean to disable this rather than introducing this dependency. This breaks abstraction layers and introduces a lot of complexity (likt the !displayAmbient condition)

MagneFire: Agreed on the hacky nature of this.
I've adjusted this so that a global boolean is used `launcherColorOverride`. It reduces the overall code size and duplication too :smile:

applauncher/003-turbo-list.qml
MagneFire: Fixed

MagneFire:
Thanks for reviewing this code!

In addition to your comments I've also renamed the `offset` and `normalizedOffset` to a more descriptive `bezelOffset` and `normalizedBezelOffset` to indicate that this has something to do with the offset with respect to the bezel.

PR #74, opened 2021-12-24
Fix issue where the background color wouldn't react to launcher changes.

MagneFire:
It turns out that when using the launcher color override the background color isn't properly switched  when switching the app launcher style (introduced: https://github.com/AsteroidOS/asteroid-launcher/pull/73).

This adds a handling for it such that this issue is solved. Not super happy with this approach as I hoped that the `onNormalizedVerOffsetChanged` function (https://github.com/AsteroidOS/asteroid-launcher/blob/3643892c263985553f7c63560948702a887599f7/qml/MainScreen.qml#L266) would take care of that. (It does, but only after switching to the home screen...)

### DIFF
--- a/applauncher/000-default-horizontal.qml
+++ b/applauncher/000-default-horizontal.qml
@@ -130,6 +130,7 @@ ListView {
         forbidBottom = false
         forbidLeft = false
         forbidRight = false
+        launcherColorOverride = false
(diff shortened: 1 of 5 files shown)


FlorentRevest:
Meh. :/ the fact that other launchers need to set it doesn’t look too good but oh well…

PR #78, opened 2022-01-15
Add previews for the stock watchfaces

eLtMosen:
Moved the previews for the stock watchfaces from the PR to unofficial.
Thanks to the hint from @MagneFire


FlorentRevest:
This is only missing 

```
watchfaces-preview.path = /usr/share/asteroid-launcher/watchfaces-preview
watchfaces-preview.files =  watchfaces-preview/*
```

Here https://github.com/AsteroidOS/asteroid-launcher/blob/master/asteroid-launcher.pro#L52

And then the rest of the series is ready to merge :)

PR #79, opened 2022-01-16
[asteroid-launcher.pro] Add watchfaces-preview path

eLtMosen:
To announce the new watchfaces-preview folder 
https://github.com/AsteroidOS/asteroid-launcher/pull/78#issuecomment-1013911078


FlorentRevest:
FYI, this is qmake jargon for "here is where to place these files when running make install" and that will make sure the asteroid-launcher openembedded recipes places the files in the right directory ;)

FlorentRevest:
It would also have been fine to do this as part of https://github.com/AsteroidOS/asteroid-launcher/pull/78 because they form one logical unit. But don't worry ahah

PR #80, opened 2022-01-23
[asteroid-launcher.pro] Add watchfaces-preview to INSTALL instruction

eLtMosen:
Correct small oversight while integrating the watchfaces-preview folder.


MagneFire:
Tested, this does indeed solve the issue.
Thanks!

PR #81, opened 2022-01-29
Add 3x3 launcher

eLtMosen:
Add 3x3 Launcher with following features
- appTitle is only shown when a gridItem is pressed for longer then 200ms
- appTitle is shown on top or bottom depending on which vertical half of the screen is pressed to mitigate "Finger hiding screen content" problem

https://user-images.githubusercontent.com/15074193/151680658-d46024a1-e18e-4147-b7cb-e144a763332d.mp4

### DIFF
--- a/applauncher/004-three-by-three.qml
+++ b/applauncher/004-three-by-three.qml
@@ -216,9 +216,7 @@ Item {
         }
 
         Component.onCompleted: {
-            launcherCenterColor = alb.centerColor(launcherModel.get(0).filePath);
-            launcherOuterColor = alb.outerColor(launcherModel.get(0).filePath);
-
+            launcherColorOverride = true
             toLeftAllowed = false
             toRightAllowed = false
             toBottomAllowed =  Qt.binding(function() { return !atYBeginning })
@@ -230,41 +228,9 @@ Item {
         }
 
         onContentYChanged: {
-
-            var lowerStop = Math.floor(contentY/appsView.height)
+            var lowerStop = Math.floor(contentY/(appsView.height/6))
             var upperStop = lowerStop+1
-            var ratio = (contentY%appsView.height)/appsView.height
-
-            if(upperStop + 1 > launcherModel.itemCount || ratio == 0) {
-                launcherCenterColor = alb.centerColor(launcherModel.get(lowerStop).filePath);
-                launcherOuterColor = alb.outerColor(launcherModel.get(lowerStop).filePath);
-                return;
-            }
-
-            if(lowerStop < 0) {
-                launcherCenterColor = alb.centerColor(launcherModel.get(0).filePath);
-                launcherOuterColor = alb.outerColor(launcherModel.get(0).filePath);
-                return;
-            }
-
-            var upperCenterColor = alb.centerColor(launcherModel.get(upperStop).filePath);
-            var lowerCenterColor = alb.centerColor(launcherModel.get(lowerStop).filePath);
-
-            launcherCenterColor = Qt.rgba(
-                        upperCenterColor.r * ratio + lowerCenterColor.r * (1-ratio),
-                        upperCenterColor.g * ratio + lowerCenterColor.g * (1-ratio),
-                        upperCenterColor.b * ratio + lowerCenterColor.b * (1-ratio)
-                    );
-
-            var upperOuterColor = alb.outerColor(launcherModel.get(upperStop).filePath);
-            var lowerOuterColor = alb.outerColor(launcherModel.get(lowerStop).filePath);
-
-            launcherOuterColor = Qt.rgba(
-                        upperOuterColor.r * ratio + lowerOuterColor.r * (1-ratio),
-                        upperOuterColor.g * ratio + lowerOuterColor.g * (1-ratio),
-                        upperOuterColor.b * ratio + lowerOuterColor.b * (1-ratio)
-                    );
-
+            var ratio = (contentY%appsView.height)/(appsView.height/6)
             currentPos = Math.round(lowerStop+ratio)
         }
     }
--- a/applauncher/004-three-by-three.qml
+++ b/applauncher/004-three-by-three.qml
@@ -51,16 +51,18 @@ Item {
 
         flow: GridView.FlowLeftToRight
         snapMode: GridView.SnapToRow
-        anchors.fill: parent
-        anchors.leftMargin: parent.width * .026
-        anchors.rightMargin: parent.width * .026
+        anchors {
+            fill: parent
+            leftMargin: parent.width * .026
+            rightMargin: parent.width * .026
+        }
         clip: true
         cellHeight: appsView.height / 3.15
         cellWidth: appsView.width / 3
-        preferredHighlightBegin: width / 3 - currentItem.width / 3
-        preferredHighlightEnd: width / 3 + currentItem.width / 3
+        preferredHighlightBegin: height / 3.15 - currentItem.height / 3.15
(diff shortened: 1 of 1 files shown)


eLtMosen:
Noticed a bug on square screens where the "shutter" would aligned to screen bottom correctly but show some offset to the bottom.

applauncher/004-three-by-three.qml
```diff
@@ -0,0 +1,358 @@
...
+                NumberAnimation { target: titleShutter; property: "opacity"; to: .85; duration: 0}
+
+                PropertyAction{}
+
+                NumberAnimation { target: hoverTitle; property: "anchors.verticalCenterOffset"; to: -hoverTitle.hoverTextOffset + (hoverTitle.hoverTextOffset * 1.62); duration: 100; easing.type: Easing.InSine}
+
+                PauseAnimation { duration: 800 }
+
+                ParallelAnimation {
+
```
FlorentRevest, line 328: Unnecessary empty line

applauncher/004-three-by-three.qml
```diff
@@ -0,0 +1,358 @@
...
+            toBottomAllowed =  Qt.binding(function() { return !atYBeginning })
+            toTopAllowed = Qt.binding(function() { return !atYEnd })
+            forbidTop = Qt.binding(function() { return !atYBeginning })
+            forbidBottom = false
+            forbidLeft = false
+            forbidRight = false
+        }
+
+        onContentYChanged: {
+            var lowerStop = Math.floor(contentY/appsView.height)
```
FlorentRevest, line 235: I don’t think this color interpolation system makes sense in a 3x3 list where there isn’t any element in focus. Wdyt?

applauncher/004-three-by-three.qml
```diff
@@ -0,0 +1,358 @@
...
+                running: false
+                repeat: false
+                onTriggered: {
+                    console.log("show label" + !dragStopper)
+                     if (!dragStopper) {
+                         parent.timedPressAndHold()
+                     }
+                     else {
+                         root.dragStop = false
+                         console.log("after IF" + !dragStopper)
```
FlorentRevest, line 128: I don’t think you meant to keep these

applauncher/004-three-by-three.qml
```diff
@@ -0,0 +1,358 @@
...
+        onDoubleClicked: mouse.accepted = false
+        onPositionChanged: mouse.accepted = false
+        onPressAndHold: mouse.accepted = false
+    }
+
+    Text {
+        id: hoverTitle
+
+        property string tempTitle: appTitle
+        property real hoverTextOffset: root.clickY > Dims.h(48) ? -Dims.h(62) : Dims.h(62)
```
FlorentRevest, line 294: Avoiding complex position calculation is best. Can’t we have a fixed height and anchors either the top or the bottom to the top or bottom of the parent ? I feel this would be a lot more readable

applauncher/004-three-by-three.qml
```diff
@@ -0,0 +1,358 @@
...
+            radius: 2.0
+            samples: 5
+            color: "#88000000"
+        }
+    }
+
+    Rectangle {
+        id: titleShutter
+
+        z: 0
```
FlorentRevest, line 350: If you can avoid manually specifying z orders by using the natural ordering of the QML file, I think it’s best.
Z orders always feel a bit hacky and hard to read, it should be avoided in situations where it can be

FlorentRevest (changes requested):

FlorentRevest:
I think this is a good candidate for upstreaming, thank you for the efforts :)

applauncher/004-three-by-three.qml
eLtMosen: I was thinking to tune it into changing to some sane color change per scrolled line of 3 icons.
I converted this to draft before @Magnefire showed how to "fix" the background color change in the 001- launcher.
So i am going to wait how we decide to handle the exact implematation there and then play with some options here before eventually deciding on no color change like you mentioned.

applauncher/004-three-by-three.qml
eLtMosen: I think i added z when i ran into an instance where items overlapped the titleShutter. But will try without now. The overlap might ave been not solved by the z even.

applauncher/004-three-by-three.qml
eLtMosen: I decided to indeed remove the flatmesh color change since changing them per line felt to much when adding the colored shutters on top. Thnaks @MagneFire for the quick `launcherColorOverride = true` hint.

applauncher/004-three-by-three.qml
```diff
@@ -0,0 +1,320 @@
...
+            Icon {
+                id: icon
+
+                width: circleWrapper.width * .70
+                height: width
+                anchors.centerIn: circleWrapper
+                color: !pressAndHoldTimer.running ?
+                           launcherItem.pressed | fakePressed ?
+                                       "#ffffff" :
+                                       "#555555" : "#555555"
```
FlorentRevest, line 210: These are overly complex ternary chains.
```
A ? B ? X : Y : Y
```
Is hard to read and should be equivalent to
```
A && B ? X : Y
```

Same comment for `circle.color`

FlorentRevest (approved):
Do you intend to keep this as a draft ? Looks good to me. This is ready to merge after the ternary conditions are addressed.

eLtMosen:
> Do you intend to keep this as a draft ? Looks good to me. This is ready to merge after the ternary conditions are addressed.

I was still refactoring the animations and at least comment the less obvious parts before putting it to review again.
Soonish ;)

applauncher/004-three-by-three.qml
eLtMosen: Due to the items being animated, it is much more complex to re-anchor them on top and bottom then the give a negative or positive offset. the code i came up with was much less readable at least and i'd like to keep using the offset. I have added comments to hopefully make the syntax easier to read.

applauncher/004-three-by-three.qml
```diff
@@ -0,0 +1,310 @@
...
+
+                PropertyAction { }
+
+                //Slide in hoverTitle and shutter to either negative or positve offset from center depending on vertical click position
+                NumberAnimation { target: hoverTitle; property: "anchors.verticalCenterOffset"; to: -hoverTitle.hoverTextOffset + (hoverTitle.hoverTextOffset * 1.58); duration: 100; easing.type: Easing.InSine}
+
+                //Keep hoverTitle in visible position for 1s
+                PauseAnimation { duration: 1000 }
+
+                //Slide hoverTitle and shutter out of view again
```
FlorentRevest, line 291: Thanks for the comments, for consistency please keep a space between the `//`s and the comments.

applauncher/004-three-by-three.qml
```diff
@@ -0,0 +1,310 @@
...
+        }
+        width: root.width
+        height: root.height * .24
+        color: alb.centerColor(launcherModel.get(currentPressedIndex).filePath)
+    }
+
+    Text {
+        id: hoverTitle
+
+        //Offset hoverText and shutter out of view in L290, either top or bottom depending on vertical click position
```
FlorentRevest, line 255: I recommend you don't reference a line number. Almost certainly someone will add or remove one line of code someday and the comment will become unreadable.

FlorentRevest (approved):
Once the two comments on comments will be addressed, feel free to merge this yourself

eLtMosen:
> Once the two comments on comments will be addressed, feel free to merge this yourself

Thanks again. I'll gladly do.

PR #82, opened 2022-01-30
Add fading transition from boot logo.

MagneFire:
This adds a transition from the bootlogo, provided by `psplash`, to the main screen/asteroid-launcher.

The Z-index is used to suppress notifications during this phase. ~This also now suppresses notifications during the tutorial meaning charging notifications are no longer interrupting the setup, making that part feel less messy too :)~
Edit: Unfortunately, we cannot simply use the `firstRun()` function for this as that would keep the main screen at a high Z-index after the tutorial since it's a function not a property (a property has a notify ability a function doesn't).

Once again, @eLtMosen helped greatly in getting this to look really nice, thanks for that!

Here are some previews:

https://user-images.githubusercontent.com/7857908/151722610-07890d33-8b9c-42a5-acd5-eef470ba62e5.mp4

https://user-images.githubusercontent.com/7857908/151722615-bb5618a2-04ef-46ec-814e-ab9bfc84f6d1.mp4

https://user-images.githubusercontent.com/7857908/151722617-c99b0b45-ba0d-4dd9-a9e8-44fae62938c6.mp4

This depends on: https://github.com/AsteroidOS/lipstick/pull/15


eLtMosen:
Just to see what is happening in the animation, here is a 3200ms long slow mo instead of the 400ms snappy version above.
https://user-images.githubusercontent.com/15074193/151723631-4f3b706c-cec4-47bd-87bf-edf301515709.mp4

qml/misc/Splash.qml
```diff
@@ -0,0 +1,102 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import QtQuick 2.9
+import org.asteroid.controls 1.0
+import QtGraphicalEffects 1.15
+
+Item {
+    anchors.fill: parent
+    visible: background.opacity > 0 ? 1.0 : 0
```
FlorentRevest, line 37: shouldn’t the possible values of the ternary operation booleans ? Aka true and false ?

FlorentRevest (approved):
Feel free to merge once the comment is addressed.

i quite like the animation good job ;)

qml/misc/Splash.qml
MagneFire: Great catch! Indeed this can be simplified wonderfully to just:
```
visible: background.opacity > 0
```

PR #83, opened 2022-02-02
[001-three-icons...] Fix app colors while scrolling

eLtMosen:
Small fix to show the correct app colors below the app starter while scrolling horizontally.
### Before, wrong behaviour:

https://user-images.githubusercontent.com/15074193/152158437-ea562496-6406-499a-99b4-4cc22653d916.mp4

### After, fixed behaviour

https://user-images.githubusercontent.com/15074193/152158559-29621956-88ce-4064-9383-baa527270b41.mp4

### DIFF
--- a/applauncher/001-three-icons-horizontal.qml
+++ b/applauncher/001-three-icons-horizontal.qml
@@ -151,8 +151,8 @@ ListView {
     }
 
     onContentXChanged: {
-        var lowerStop = Math.floor(contentX/appsListView.width)
-        var upperStop = lowerStop+1
+        var lowerStop = Math.floor((contentX * 2)/appsListView.width)
+        var upperStop = lowerStop + 1
         var ratio = (contentX%appsListView.width)/appsListView.width
 
         if(upperStop + 1 > launcherModel.itemCount || ratio == 0) {
--- a/applauncher/001-three-icons-horizontal.qml
+++ b/applauncher/001-three-icons-horizontal.qml
@@ -151,21 +151,10 @@ ListView {
     }
 
     onContentXChanged: {
-        var lowerStop = Math.floor((contentX * 2)/appsListView.width)
+        var w = (appsListView.width / 2)
+        var lowerStop = Math.floor((contentX + w/2) / w)
         var upperStop = lowerStop + 1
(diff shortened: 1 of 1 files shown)


applauncher/001-three-icons-horizontal.qml
```diff
@@ -151,21 +151,10 @@ ListView {
     }
 
     onContentXChanged: {
-        var lowerStop = Math.floor(contentX/appsListView.width)
-        var upperStop = lowerStop+1
-        var ratio = (contentX%appsListView.width)/appsListView.width
-
-        if(upperStop + 1 > launcherModel.itemCount || ratio == 0) {
```
FlorentRevest, line 158: I’m surprised you could remove these two if blocks.

I suppose this launcher can also « overshoot » past the first or last icon, right ? Then I think your color interpolation formula doesn’t work in these corner cases and lowerStop or upperStop would be invalid indices, wouldn’t they ?

eLtMosen: *Points fingers* That was @MagneFire idea /o\
Kidding. I do not fully understand tbh. I was sure to have tested thoroughly, after those stops where removed. Especially what you (and i from now on) call overshoot. Since from my changes in line 150 overshoot caused the wrong colors to be assigned when scrolling the list back at some point. Magnefire corrected that and removed those two ifs in that process. My testing started after that and i found no apparent issues.
Tested with various amounts of apps installed on different watches.
Could you elaborate on what the exact implication is. I.e what to look for in testing?
Have you tested on device?

FlorentRevest: I have not tested on device. For example let’s say you have two items in the list and you slide fast all the way to the right such that the list overshoots a bit. Then lowerStop will be 1 (second item) and upperStop 2 (third item in the list, which does not exist) so when getting launcherModel.get(upperStop) you would not get a color and I expect the function would fail.

Similarly, if you were to slide all the way to the left, the overshoot would produce negative contentX so lowerStop would be -1 or something which would fail similarly.

these if blocks made sure that in the boundary conditions overshoot would just keep the first/last item’s colors

eLtMosen:
Understood the possible implications now and agree to keep the lower and upperStop.

PR #86, opened 2022-02-20
[000-default-horizontal] Enhance legibility

eLtMosen:
Due to the app-color-theme update and addition of several new launchers, it was discussed that the stock launcher now appears to have too large circleWrapper in comparison. The new app colors would be more noticeable when the circle gets smaller and covers less of the background.

- Change circle size from 0.7 to 0.65 parent size
- Adapt font top margin accordingly
- Select SemiBold font style for better legibility due to more weight
- Darken icon and text from #444 to #333
- Change Icon size to .3 from 3.1 to reduce anti aliasing due to an even divider. 
- Make circle transparent instead of darker to slightly color it in the app color
- Make circle Dropshadow slightly darker
- Add 100ms transition anim to onPressed toggle

One benefit i observed is slightly less screen tearing while scrolling on sparrow.

Before/After

https://user-images.githubusercontent.com/15074193/154825058-6f9358ed-d064-4c57-ae9e-9a931f65f935.mp4

onPressed Before/After

https://user-images.githubusercontent.com/15074193/154825078-466b505a-9720-465e-bf40-dd77f9d13e74.mp4

Scroll behviour

https://user-images.githubusercontent.com/15074193/154825193-1faabdb5-6be4-47e1-8447-05765c0ee271.mp4

### DIFF
--- a/applauncher/000-default-horizontal.qml
+++ b/applauncher/000-default-horizontal.qml
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Timo Könnecke <github.com/eLtMosen>
+ * Copyright (C) 2022 Timo Könnecke <github.com/eLtMosen>
  *               2015 Florent Revest <revestflo@gmail.com>
  *               2014 Aleksi Suomalainen <suomalainen.aleksi@gmail.com>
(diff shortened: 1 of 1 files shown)


FlorentRevest:
Thank you, this looks great! :D

PR #87, opened 2022-03-16
Hide wallpaper in AOD mode, disable home screen when app is visible.

MagneFire:
This allows for the disabling of qml loaded components, potentially reducing battery consumption in ambient mode.

Additionally, this disables the home screen when an app is visible, improving performance in apps by a frame or two per second as well :)

### DIFF
--- a/qml/MainScreen.qml
+++ b/qml/MainScreen.qml
@@ -367,6 +367,8 @@ Item {
         id: wallpaper
         width: Dims.w(110)
         height:  Dims.h(110)
+        visible: wallpaperDarkener.opacity != 1.0
+        enabled: visible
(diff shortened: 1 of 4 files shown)


MagneFire:
Last commit introduces a flicker when opening apps (home screen -> black screen -> app). Please wait before this is fixed.

MagneFire:
The issue has been solved by introducing a `aboutToOpen` variable.

FlorentRevest:
Nice! :) This was indeed suggested by https://resources.qt.io/videos/qtws17-intro-to-the-webos-qtwayland-compositor-florian-haenel-lg-electronics

PR #89, opened 2022-07-12
Implement nightstand mode

beroset:
This corresponds to https://github.com/AsteroidOS/asteroid/issues/151
and https://github.com/AsteroidOS/asteroid-settings/issues/48 and is the
complementary change required to complete the implementation of
nightstand mode.

When the user has nightstand mode enabled, when a charging cable is
detected, the display changes to the user-selected watchface and
brightness for that mode.  It also starts a timer with the delay
interval set by the user.  If the charging cable is disconnected, the
timer will wait for at least that delay before changing back to the
daytime watchface.  Note that in ambient mode, the timing is suspended
for intervals so the delay may be longer than set by the user.

Signed-off-by: Ed Beroset <beroset@ieee.org>

### DIFF
--- a/qml/MainScreen.qml
+++ b/qml/MainScreen.qml
@@ -31,7 +31,9 @@
 import QtQuick 2.9
 import Nemo.Time 1.0
 import Nemo.Configuration 1.0
+import Nemo.Mce 1.0
 import org.nemomobile.lipstick 0.1
+import org.nemomobile.systemsettings 1.0
 import org.asteroid.controls 1.0
 import org.asteroid.utils 1.0
 import org.asteroid.launcher 1.0
@@ -156,6 +158,10 @@ Item {
         updateFrequency: WallClock.Second
     }
 
+    MceCableState {
+        id: mceCableState
+    }
+
     ConfigurationValue {
         id: use12H
         key: "/org/asteroidos/settings/use-12h-format"
@@ -171,6 +177,10 @@ Item {
 
     LauncherModel {
         id: launcherModel
+    }
+
+    DisplaySettings { 
+        id: displaySettings
     }
 
     Connections {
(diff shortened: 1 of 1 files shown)


qml/MainScreen.qml
```diff
@@ -217,6 +227,56 @@ Item {
...
+    }
+
+    ConfigurationValue {
+        id: nightstandEnabled
+        key: "/desktop/asteroid/nightstand/enabled"
+        defaultValue: false
+    }
+
+    Item {
+        id: nightstandMode
```
FlorentRevest, line 255: Nice encapsulation! :) We should do this more often

qml/MainScreen.qml
```diff
@@ -217,6 +227,56 @@ Item {
...
+    ConfigurationValue {
+        id: nightstandBrightness
+        key: "/desktop/asteroid/nightstand/brightness"
+        defaultValue: 30
+    }
+
+    ConfigurationValue {
+        id: nightstandDelay
+        key: "/desktop/asteroid/nightstand/delay"
+        defaultValue: 5
```
FlorentRevest, line 245: Would you mind documenting above what's the purpose of this function (not what it does but more, why you felt the need to introduce it) and why you came up with 5 seconds as a default value?

Spontaneously I seem to think of it as a workaround for bad connections and 5 seconds seems a bit high, maybe 2 or 3 seconds would have been enough but maybe I'm missing something ? Because I imagine it must feel very "laggy" if disconnecting the watch from its charger keeps this on for 5 seconds

qml/MainScreen.qml
```diff
@@ -217,6 +227,56 @@ Item {
...
+
+    ConfigurationValue {
+        id: nightstandEnabled
+        key: "/desktop/asteroid/nightstand/enabled"
+        defaultValue: false
+    }
+
+    Item {
+        id: nightstandMode
+        readonly property bool active: ready || (nightstandEnabled.value && holdoff)
```
FlorentRevest, line 256: Am I right that if you remove `ready || (nightstandEnabled.value &&` here the behavior would not change?

There shouldn't be any situation when `ready` is true and `holdoff` is false because `holdoff` gets set to true everytime `ready` is set to true and it only transitions to false a while after `ready` goes back to false. So the `ready ||` part seems redundant ?

Also there shouldn't be any situation when `holdoff` is true and `nightstandEnabled.value` is false because `holdoff` is only set to true when ready transitions to true which can only happens if `nightstandEnabled.value` is true...

So how about removing this definition of `ready` and the holdoff variable and write to ready  instead of write to holdoff ?

qml/MainScreen.qml
```diff
@@ -217,6 +227,56 @@ Item {
...
+        readonly property bool ready: nightstandEnabled.value && mceCableState.connected
+        property bool holdoff: false
+        property int oldBrightness: 100
+        onReadyChanged: {
+            if (ready) {
```
FlorentRevest, line 261: Shouldn't you cancel `nightstandDelayTimer` here ?

Let's say you plug your watch in but it has a bad contact so it:

- gets a connection (ready transitions to true)
- immediately afterward gets a disconnection (ready transitions to false and triggers nightstandDelayTimer)
- gets a connection (ready transitions back to true)

Now the watch stays in that state but the timer elapses and disabled the timer.

---

I think this is maybe the bug you camouflaged behind complicated boolean expressions but with a simplified expression and a more sane logic here this would all work out ?

qml/MainScreen.qml
```diff
@@ -217,6 +227,56 @@ Item {
...
+                oldBrightness = displaySettings.brightness
+                displaySettings.brightness = nightstandBrightness.value
+                holdoff = true
+            } else {
+                displaySettings.brightness = oldBrightness
```
FlorentRevest, line 266: Shouldn't the original brightness be restored only when active transitions back to false ? that is when the timer elapses ?

qml/MainScreen.qml
beroset: If it's 3AM and I grab the watch in the dark without my glasses on to take a look, I don't want it to immediately switch to daytime watchface, but maintain the more readable nightstand watchface for a few seconds.

beroset: See also https://github.com/AsteroidOS/asteroid/issues/151#issuecomment-1204035843 for an elaboration of that scenario and corresponding requirements.

qml/MainScreen.qml
beroset: The logic was a bit convoluted, but now it implements this state machine using only two boolean properties.  New code comments also should help explain.
transitions

qml/MainScreen.qml
beroset: You're right.  I think the current implementation is much cleaner.

qml/MainScreen.qml
beroset: There are two circumstances in which it should immediately restore the original brightness:

1. when the timer expires
2. if nightstand mode is disabled

The new version of the code does exactly this.

beroset:
I think I have everything resolved now, and it's a bit more straightforward internal logic as well.

qml/MainScreen.qml
```diff
@@ -66,6 +68,8 @@ Item {
     property bool aboutToClose: false
     property bool aboutToMinimize: false
 
+    property alias nightstand: nightstandMode.active
```
FlorentRevest, line 71: IIuc this is for watchfaces right? Could you add a comment above to explain why it's not used in this file and that it's meant to be a stable API for watchface developers ?

FlorentRevest (approved):
Ready to merge with the one comment addressed and the corresponding asteroid-settings PR merged :)

qml/MainScreen.qml
beroset: I've added a comment to that effect.

PR #91, opened 2022-07-28
Fix #90 by making bluetooth agent round

### LINKED ISSUE #90: Bluetooth Dialog is misshapen on beluga, smelt, carp and minnow
For all watches where the screen isn't 1:1 proportioned, this element turns into an elongated rounded rectangle.

dodoradio:
A very blunt fix for #90 - bluetooth dialog not being round on watches with non-square displays. There might be other elements with a similar issue, but I haven't run into them yet.

### DIFF
--- a/qml/misc/BluetoothAgent.qml
+++ b/qml/misc/BluetoothAgent.qml
@@ -36,7 +36,7 @@ Item {
     id: btAgent
 
     width: initialSize.width
-    height: initialSize.height
+    height: width
 
     Item {
         id: circleWrapper


FlorentRevest (approved):

FlorentRevest:
To be entirely future-proof, we should also use `width: min(initialSize.width, initialSize.height)` because for example minnow is wider than it is tall and this would still cause notifications whose proportions are off.

It was kind of the idea between `Dims.w` (width percentage) `Dims.h` (height percentage) and `Dims.l` (percentage of the shortest dimension of the screen) so we can draw circles of the dimension that fits in the screen. So maybe `width: Dims.l(100)` helps here ?

But to be fair I'm not sure if the `btAgent` top item is even centered (probably not?) so it sounds like changing the size of the top item will make the notification round but it will be positioned from the top left corner and lead to weird behaviors still ?

FlorentRevest:
Merging this anyway because it's an improvement over the status quo but thanks for taking a look if the comments above are actual issues

PR #100, opened 2022-10-23
010 add chargering

eLtMosen:
- Add nightstand charge Arc

### DIFF
--- a/watchfaces/010-masked-spartan.qml
+++ b/watchfaces/010-masked-spartan.qml
@@ -1,5 +1,7 @@
 /*
- * Copyright (C) 2018 - Timo Könnecke <el-t-mo@arcor.de>
+ * Copyright (C) 2022 - Timo Könnecke <github.com/eLtMosen>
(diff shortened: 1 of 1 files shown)


watchfaces/010-masked-spartan.qml
```diff
@@ -170,4 +189,60 @@ Item {
                 "
         }
     }
+
+    Item {
+        id: nightstandMode
+
+        readonly property bool active: mceCableState.connected //ready || (nightstandEnabled.value && holdoff)
+        //readonly property bool ready: nightstandEnabled.value && mceCableState.connected
```
MagneFire, line 197: This looks like an error to me?

eLtMosen: Oh wow, great catch. That is an oversight and should have been chnaged to `nightstand` like in the other WFs.

eLtMosen:
007-resized

Resized the content during nightstand so a sane burnin offset is possible

FlorentRevest (approved):

PR #110, opened 2022-10-23
000 add chargering

eLtMosen:
- Add segmented nightstand charge Arc
- Resize content to make room for charge Arc
- Optimize burnin offset

### DIFF
--- a/watchfaces/000-default-digital.qml
+++ b/watchfaces/000-default-digital.qml
@@ -38,12 +38,9 @@ import org.asteroid.utils 1.0
 import Nemo.Mce 1.0
 
 Item {
-    id: rootitem
+    id: root
 
-    width: parent.width * (nightstandMode.active ? .9 : 1)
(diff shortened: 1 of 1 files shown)


watchfaces/000-default-digital.qml
```diff
@@ -120,6 +126,102 @@ Item {
         }
     }
 
+    Item {
+        id: nightstandMode
+
+        readonly property bool active: mceCableState.connected //ready || (nightstandEnabled.value && holdoff)
```
FlorentRevest: Isn't that supposed to be provided by asteroid-launcher itself ?

eLtMosen: Ugh, indeed. Same issue as @MagneFire spotted in 010-.
I was not careful enough checking all those PRs :P

FlorentRevest (approved):

PR #113, opened 2022-11-25
notification sound effect

lepras:
in reference to [this](https://log.asteroidos.org/?date=2022-11-25&level=10#104) chat on matrix.

### DIFF
--- a/qml/notifications/NotificationView.qml
+++ b/qml/notifications/NotificationView.qml
@@ -29,6 +29,7 @@
 
 import QtQuick 2.9
 import org.asteroid.controls 1.0
+import QtMultimedia 5.4
 
 MouseArea {
     id: view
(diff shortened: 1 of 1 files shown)


qml/quicksettings/QuickSettings.qml
```diff
@@ -63,13 +66,13 @@ Item {
     }
 
     QuickSettingsToggle {
-        id: lockedToggle
+        id: soundToggle
```
MagneFire, line 69: These changes should not be part of this PR.

We should discuss in a separate issue. Obviously it would be nice to have a quick button to mute all sounds, but not every watch has a speaker rendering this button entirely useless.
It also replaces an already existing function, which might not be desirable right now.

MagneFire (changes requested):
Thank you for this PR! And I'm sorry for taking this long to review it.

Please see the feedback on this PR.

When you remove the QuickSettings changes I think this PR should be ready to merge.
However, I think it would be better to eventually add these sound related events to the `ngfd` events: https://github.com/AsteroidOS/meta-asteroid/blob/master/recipes-nemomobile/ngfd/ngfd/events.d/alarm.ini#L1 and add some of the lines references here: https://github.com/sailfishos/ngfd/blob/master/data/events.d/clock.ini#L2.
We could even take advantage of profiles to alter the behaviour of alarm events: https://github.com/sailfishos/ngfd/blob/master/data/events.d/ringtone.ini#L4 this would remove the need to dynamically change the system volume.

qml/quicksettings/QuickSettings.qml
lepras: removed that commit

lepras:
re-synced to the upstream for merging.

MagneFire (approved):

PR #114, opened 2022-11-26
MainScreen: Fix issue where DeviceInfo is defined in multiple QML modules.

MagneFire:
Use qualifiers for org.nemomobile.systemsettings and org.asteroid.utils to fix unpredictable behaviour with the DeviceInfo as it's defined by both modules. This fixes an issue where burn-in-protection would remain enabled eventhough the user disabled it in the settings app.

This fixes: https://github.com/AsteroidOS/meta-smartwatch/issues/133

### DIFF
--- a/qml/MainScreen.qml
+++ b/qml/MainScreen.qml
@@ -33,9 +33,9 @@ import Nemo.Time 1.0
 import Nemo.Configuration 1.0
 import Nemo.Mce 1.0
 import org.nemomobile.lipstick 0.1
-import org.nemomobile.systemsettings 1.0
+import org.nemomobile.systemsettings 1.0 as NemoSystemSettings
 import org.asteroid.controls 1.0
-import org.asteroid.utils 1.0
+import org.asteroid.utils 1.0 as AsteroidUtils
 import org.asteroid.launcher 1.0
 import "desktop.js" as Desktop
 
@@ -104,7 +104,7 @@ Item {
     ConfigurationValue {
         id: useBip
         key: "/org/asteroidos/settings/use-burn-in-protection"
-        defaultValue: DeviceInfo.needsBurnInProtection
+        defaultValue: AsteroidUtils.DeviceInfo.needsBurnInProtection
     }
 
     Item {
@@ -124,7 +124,7 @@ Item {
         property int heightOffset
 
         // Enable/disable burn in protection.
-        enabled: DeviceInfo.needsBurnInProtection && useBip.value
+        enabled: AsteroidUtils.DeviceInfo.needsBurnInProtection && useBip.value
 
         onHeightOffsetChanged: {
             topOffset = heightOffset/2
@@ -186,7 +186,7 @@ Item {
         id: launcherModel
     }
 
-    DisplaySettings { 
+    NemoSystemSettings.DisplaySettings { 
         id: displaySettings
     }
 


FlorentRevest (approved):
Ugh... It's pretty terrible if a qml files has to do things differently just because another qml file it doesn't know about imports the same module...
This kind of looks like a Qt bug to me actually. I'd expect that DeviceInfo is a singleton anyway and is always the same object. There shouldn't be anything unpredictable. I'd be curious to see if this is fixed in newer Qt or if this is something the Qt team is aware of.

Anyway... as a temporary workaround, let's go with this. I think that we have other hacks in asteroid-settings already.

MagneFire:
> Ugh... It's pretty terrible if a qml files has to do things differently just because another qml file it doesn't know about imports the same module... This kind of looks like a Qt bug to me actually. I'd expect that DeviceInfo is a singleton anyway and is always the same object. There shouldn't be anything unpredictable. I'd be curious to see if this is fixed in newer Qt or if this is something the Qt team is aware of.
> 
> Anyway... as a temporary workaround, let's go with this. I think that we have other hacks in asteroid-settings already.

Well, two modules export a `DeviceInfo` object. I'd expect some kind of runtime error instead of unpredictable behavior. The implemented workaround is at least an old method: http://imaginativethinking.ca/resolve-component-naming-conflicts-qml/ (2014).

It's indeed weird that this worked for how long it did. I did consider using the Qualifier (https://doc.qt.io/qt-6/qtqml-syntax-imports.html#module-namespace-imports) for all imports, but I don't think that makes sense either.

FlorentRevest:
> Well, two modules export a `DeviceInfo` object. I'd expect some kind of runtime error instead of unpredictable behavior. The implemented workaround is at least an old method: http://imaginativethinking.ca/resolve-component-naming-conflicts-qml/ (2014).

But DeviceInfo is a singleton so it shouldn't make a difference which DeviceInfo is used anyway.

I wonder if we should avoid exposing singletons and just having a DeviceInfo class which can be instantiated and which, under the hood, would get its values from a singleton.

MagneFire:
> > Well, two modules export a `DeviceInfo` object. I'd expect some kind of runtime error instead of unpredictable behavior. The implemented workaround is at least an old method: http://imaginativethinking.ca/resolve-component-naming-conflicts-qml/ (2014).
> 
> But DeviceInfo is a singleton so it shouldn't make a difference which DeviceInfo is used anyway.
> 
> I wonder if we should avoid exposing singletons and just having a DeviceInfo class which can be instantiated and which, under the hood, would get its values from a singleton.

True, but in that case it might matter when the singleton is actually loaded. `MainScreen` is the root element of the homescreen application.
It might be interesting to test the following case:
- Root application (main.qml) that only imports `org.asteroid.utils` and uses `DeviceInfo`.
- A sub page (xx.qml) that imports both `org.nemomobile.systemsettings` and `org.asteroid.utils`. Now we expect the `DeviceInfo` from  `org.asteroid.utils` to be used.

Actually that might be exactly what the settings app is doing. In which case an easier solution could've been to maybe swap the import ordering (although that's a bit of a weird hack as well)?

FlorentRevest:
> > > Well, two modules export a `DeviceInfo` object. I'd expect some kind of runtime error instead of unpredictable behavior. The implemented workaround is at least an old method: http://imaginativethinking.ca/resolve-component-naming-conflicts-qml/ (2014).
> > 
> > 
> > But DeviceInfo is a singleton so it shouldn't make a difference which DeviceInfo is used anyway.
> > I wonder if we should avoid exposing singletons and just having a DeviceInfo class which can be instantiated and which, under the hood, would get its values from a singleton.
> 
> True, but in that case it might matter when the singleton is actually loaded. `MainScreen` is the root element of the homescreen application. It might be interesting to test the following case:
> 
> * Root application (main.qml) that only imports `org.asteroid.utils` and uses `DeviceInfo`.
> * A sub page (xx.qml) that imports both `org.nemomobile.systemsettings` and `org.asteroid.utils`. Now we expect the `DeviceInfo` from  `org.asteroid.utils` to be used.
> 
> Actually that might be exactly what the settings app is doing. In which case an easier solution could've been to maybe swap the import ordering (although that's a bit of a weird hack as well)?

Yeah my point is that we shouldn't have to resort to this kind of dark magic to get things to work. A well designed language should work like you can reasonably expect.

Here if QML gets in the way (that's the first time) maybe we can work around it by modifying qml-asteroid to force the user into adopting a safer practice. This will avoid that app developers inadvertently shoot themselves in the foot.

PR #119, opened 2023-01-10
watchfaces: Fix issue where the month isn't updated.

MagneFire:
The getDate() function returns the day of the month ([^1]) so it will never change when the month is changed. Which can naturally happen when the month is changed in the settings app or synchonized. This adds the usage of getMonth() to the relevant watchfaces to fix this issue.

[^1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDate

(This is a draft PR until I've actually tested this completely)

### DIFF
--- a/watchfaces/000-default-digital.qml
+++ b/watchfaces/000-default-digital.qml
@@ -127,6 +127,7 @@ Item {
             id: dateCanvas
 
             property int date: 0
+            property int month: 0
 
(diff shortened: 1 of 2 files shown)


eLtMosen:
Tested and works for me®

eLtMosen (approved):
Tested and approved

PR #123, opened 2023-03-27
Fix 3x3 launcher margins and remove dead code

### LINKED ISSUE #122: [3x3 launcher] Bottom end of the listview is offset
- the scroll end of the list view should stop "one icon row" earlier so that only the last of 3 rows is shown empty and overshoots to only one line of icons spacing
- The bottom indicator (small white diamond on the bottom) should only be shown much later

screenshot-3x3-1

Currenlty, scrolling stops with two empty icon rows on the bottom like shown above.
And overshoots to only one icon row visible

screenshot-3x3-2

beroset:
This patch does a few things intended to fix issues and clean up the code a bit.

First, instead of using preferredHighlightBegin and preferredHighlightEnd and highlightRangeMode, we simply use header and footer and point to a square Item component that is half the size of a row.  This fixes #122 .

Second, some dead code was removed: two unused Connections were removed, and the launcherColorOverride setting.

Third, the delegate for the grid was changed to a named Component and moved from inline.  This helps readability. See
https://doc.qt.io/qt-6/qml-qtquick-gridview.html#example-usage for the description of "An improved grid view" where a similar change was made in the example.

### DIFF
--- a/applauncher/004-three-by-three.qml
+++ b/applauncher/004-three-by-three.qml
@@ -1,5 +1,6 @@
 /*
- * Copyright (C) 2022 Timo Könnecke <github.com/eLtMosen>
+ * Copyright (C) 2023 Ed Beroset <beroset@ieee.org>
+ *               2022 Timo Könnecke <github.com/eLtMosen>
  *               2021 Darrel Griët <dgriet@gmail.com>
  *               2015 Florent Revest <revestflo@gmail.com>
  *               2014 Aleksi Suomalainen <suomalainen.aleksi@gmail.com>
@@ -35,6 +36,7 @@
 import QtQuick 2.15
 import QtGraphicalEffects 1.12
 import org.asteroid.controls 1.0
+import org.asteroid.utils 1.0 as AsteroidUtils
 
 Item {
     id: root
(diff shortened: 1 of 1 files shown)


applauncher/004-three-by-three.qml
```diff
@@ -204,9 +145,60 @@ Item {
...
+                grid.changeAllowedDirections()
+            }
+        }
+
+        model: launcherModel
+
+        delegate: gridDelegate
 
         Component.onCompleted: {
-            launcherColorOverride = true
```
FlorentRevest, line 209: This is not dead code, this is affecting the behavior of https://github.com/AsteroidOS/asteroid-launcher/blob/d5f4f7361b9896fedbc03e575c2576b6fac74bab/qml/MainScreen.qml#L364

I know... This is a pretty terrible pattern to have these implicit globally defined variables :( Suggestions welcome

applauncher/004-three-by-three.qml
```diff
@@ -204,9 +145,60 @@ Item {
                 }
             }
         }
+    }
+
+    Component {
+        id: spacer
+        Item {
+            height: root.height / numColumns / 2
```
FlorentRevest, line 153: While we are at it, let's make this height conditional on whether the screen is round or not.

Design decision ratified by @eLtMosen :p

applauncher/004-three-by-three.qml
```diff
@@ -150,13 +98,6 @@ Item {
                 currentPressedIndex = index
             }
 
-            Connections {
-                target: appLauncher
-                function onFakePressedChanged() {
```
FlorentRevest, line 155: This fake pressed variable is also an implicit global variable, it is used to implement the tutorial.

I guess we should question whether any launcher apart from the 000 default launcher should have logic for the tutorial... But let's keep separately

applauncher/004-three-by-three.qml
```diff
@@ -45,68 +46,15 @@ Item {
...
-            bottomIndicator.animate()
-        }
-
-        onDragStarted: {
-            dragStop = true
-        }
-
-        Connections {
-            target: grid
-            function onCurrentVerticalPosChanged() {
```
FlorentRevest, line 82: This is also linked to a global variable and makes sure that if you: open an app, let the watch go to sleep, unlock, go back to the launcher, you are at the top of the scroll view rather than halfway through it. It should also be kept

applauncher/004-three-by-three.qml
FlorentRevest: Or maybe we should have a similar margin as the GridView (`parent.width * .026` ?) idk, whatever looks best.

applauncher/004-three-by-three.qml
```diff
@@ -45,68 +46,15 @@ Item {
...
-            fill: parent
-            leftMargin: parent.width * .026
-            rightMargin: parent.width * .026
-        }
-        clip: true
-        cellHeight: appsView.height / 3.15
-        cellWidth: appsView.width / 3
-        preferredHighlightBegin: height / 3.15 - currentItem.height / 3.15
-        preferredHighlightEnd: height / 3.15 + currentItem.height / 3.15
-        highlightRangeMode: ListView.StrictlyEnforceRange
```
FlorentRevest, line 64: Somehow I have the feeling that these fields were trying to solve something else than the footer/header problem you've solved, yet, I can't quite put my finger on what that would be.

@MagneFire it seems that you wrote this code, maybe you remember what you were trying to achieve here ?

MagneFire: Originally, this logic was used to ensure an equal amount of spacing on the top and bottom of the `GridView`. In this launcher it was used to shift the grid a bit further to the top resulting in the weird bug where too much padding is available at the bottom.
Removing this and adding header and footer components is the right way to go here.

applauncher/004-three-by-three.qml
MagneFire: Can confirm that the 'go to home gesture' leaves the app launcher at a the previous location, which isn't intended for these vertical app launchers.

MagneFire (commented):
Agreed on the comments by @FlorentRevest.
Other than that, this padding fix makes this launcher feel much more refined, thanks for that!

applauncher/004-three-by-three.qml
eLtMosen: > While we are at it, let's make this height conditional on whether the screen is round or not.
> 
> Design decision ratified by @eLtMosen :p

Indeed, i was only made aware by @FlorentRevest that it might make sense to have no bottom margin on square screens and thus to show all 3 rows. Fully agree on that and still question myself for not having noticed while testing on sparrow. `/o\`

applauncher/004-three-by-three.qml
beroset: I'll add a comment to note variables which are from the parent context.  There's probably a better way to do it, but I can't think of it at the moment.

beroset: Also, when selecting a launcher, this variable is not available from within `asteroid-settings` so the result is an error message (for each launcher):   
```
Mar 28 10:06:55 catfish asteroid-launcher[2170]: file:///usr/share/asteroid-launcher/applauncher/004-three-by-three.qml:216: Error: Invalid write to global property "launcherColorOverride"
```
But that's an issue for `asteroid-settings` so I will add another PR there.

applauncher/004-three-by-three.qml
beroset: Done!  I tried it on `dory` (my only square watch) and it looks good.

applauncher/004-three-by-three.qml
beroset: Thanks for the confirmation.  I read about these fields and came to that conclusion, but it's nice to get feedback.

applauncher/004-three-by-three.qml
beroset: I have restored it, but the behavior I observe is the same with or without it, so I'll leave it to someone else to resolve.  Philosophically, I don't know why the horizontally organized launchers keep their place (which I find convenient) but the vertically organized ones should not.

applauncher/004-three-by-three.qml
beroset: Restored.

FlorentRevest:
Lovely, thank you! :D one less thing holding back 2.0 🚀

PR #131, opened 2023-04-09
Remove analog-silver-swerver

eLtMosen:
Due to sever bug that shows boxes instead of the numerals


FlorentRevest:
Uh that's kind of awkward we remove high level features due to bugs in our graphic stack...

Let's do this for now since we still have that watchface in the unofficial-watchfaces repo but fundamentally we should try to isolate a reproducer for the bug and have it go through a wizard like @MagneFire to get to the bottom of things and solve it for real.

FlorentRevest (approved):

Issue #133, opened 2023-04-16
Add newly promoted stock watchfaces


eLtMosen:
Add analog-words https://github.com/AsteroidOS/asteroid-launcher/pull/129
Add analog-weather-satellite https://github.com/AsteroidOS/asteroid-launcher/pull/128
Add numerals-duo-synth-neon-green https://github.com/AsteroidOS/asteroid-launcher/pull/127
Add analog-aviator https://github.com/AsteroidOS/asteroid-launcher/pull/126
Add analog-scientific-v2 https://github.com/AsteroidOS/asteroid-launcher/pull/125
Add analog-circle-shades https://github.com/AsteroidOS/asteroid-launcher/pull/124

Reordering is discussed here
https://github.com/AsteroidOS/asteroid-launcher/issues/132

Issue #132, opened 2023-04-12
Reorder stock watchfaces after adding 6 new ones


eLtMosen:
Currently they will just be added at the end. It might be nice to sort them visually appealing.

eLtMosen:
My current thinking after shifting the W around in gimp would be like so
- Keep the analog/digital left/right toggle per row we had in old sort order
- Sort the less clean ones down
- Add digital-outfit as filler to have more digital faces for the analog/digital toggle per row and have 18 WF in all, filling all rows.

Imho digital-outfit makes sense to add since we have very few clean digital faces in stock. It is a bit redundant with the 003-alternative-digital-2. But still is different enough due to more elegant font style and even more simplistic layout to justify having both of these designs in stock.
Another possible candidate for a filler would be digital-shifted. That is a bit too funky and extreme for stock tho. Another candidate would be digital-thoreau-simplify. Which might need some polishing for stock addition imo.

![grafik](https://user-images.githubusercontent.com/15074193/232321757-9a173d65-42f3-4dbf-8a0d-a078e7cb24f8.png)

PR #135, opened 2023-04-19
Reorder watchfaces after adding 7 new ones after promotion

eLtMosen:
- Add digital-outfit as filler to have more digital watchfaces to keep the zig/zag toggle per row
- Sort and thus rename watchfaces according to discussion in this issue https://github.com/AsteroidOS/asteroid-launcher/issues/132

Fixes https://github.com/AsteroidOS/asteroid-launcher/issues/132
Fixes https://github.com/AsteroidOS/asteroid-launcher/issues/133

grafik


MagneFire (approved):
LGTM!

PR #138, opened 2023-04-29
applauncher: speed dial: Move pill to the left & crown support.

MagneFire:
The pill movement has two benefits:
- Your finger is less likely to overlap the app name label (pointed out by @eLtMosen).
- App icons are aligned with the rotating crown (often located on the right).

Here's it in action on `hoki`:

https://user-images.githubusercontent.com/7857908/235305799-70bcd7f6-4603-4426-a0df-2bb12c590ca6.mp4

### DIFF
--- a/applauncher/005-speed-dial.qml
+++ b/applauncher/005-speed-dial.qml
@@ -215,6 +215,24 @@ Item {
                 grid.changeAllowedDirections()
             }
         }
+
+        WheelHandler {
+            property: "offset"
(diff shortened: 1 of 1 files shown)


eLtMosen (approved):
Compiled and tested on narwhal.
Its astonishing how much sense the mirroring makes.
As pointed out, the Title text looks much better when the Capital letter starts at the left side and ends with a lower case character into the pill rounding. Its a spacing thing and look much better now when mirrored

PR #161, opened 2023-06-03
Fix off-by-one error in agenda display

beroset:
As noted by aranor.16 on Matrix chat, the watch is supposed to show today's agenda items when you swipe left, but there was an off-by-one error that only showed items for the next month on the same day.  This has been fixed, and it now works perfectly with the calendar.


MagneFire:
Thanks for the quick fix!

PR #166, opened 2023-09-04
Fix check for no weather data

beroset:
This modifies the weather data to check for 0K instead of trying to convert before checking. Now days for which there is no weather data clearly show "No Weather Data" instead of showing -459F regardless of whether the user has chosen C or F for their preferred temperature units.


eLtMosen (approved):
yay, thanks for the swift action!

PR #167, opened 2023-09-12
GestureFilterArea: Use local positioning for movement detection.

MagneFire:
This is instead of the window position.

It may occur that the GestureFilterArea is used in a nested way, where a higher level item can be rotated. This change will allow the GestureFilterArea to work as intended when it's rotated/transformed/scaled.

It essentially removes the need for this patch: https://github.com/AsteroidOS/meta-smartwatch/blob/cc4fbcf41bfe83ddba4128d54086acab6b58d65b/meta-mtk6580/recipes-asteroid/asteroid-launcher/asteroid-launcher/0002-GestureFilterArea-Fix-swipe-interaction-on-rotated-s.patch.

Once this is merged a change to https://github.com/AsteroidOS/meta-smartwatch/blob/cc4fbcf41bfe83ddba4128d54086acab6b58d65b/meta-mtk6580/recipes-asteroid/asteroid-launcher/asteroid-launcher_%25.bbappend is also required.

### DIFF
--- a/src/gesturefilterarea.cpp
+++ b/src/gesturefilterarea.cpp
@@ -82,7 +82,7 @@ void GestureFilterArea::mousePressEvent(QMouseEvent *event) {
         m_velocityY = 0;
         m_tracing = true;
         m_horizontal = false;
-        m_prevPos = event->windowPos();
+        m_prevPos = event->localPos();
         m_counter = 0;
     }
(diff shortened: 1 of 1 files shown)


FlorentRevest:
Good catch! Much cleaner indeed

PR #168, opened 2023-10-02
MainScreen: Use stock watchface for nightstand mode.

MagneFire:
The analog-nordic watchface was moved to the community watchfaces, this meant that errors would be printed when the launcher is started.
Switch back to the stock watchface as we now have charging rings for all stock watchfaces in nightstand mode.


beroset (approved):
Good change.  Thanks!

PR #171, opened 2024-05-16
Make 3x3 launcher reset when view changes to watchface

dodoradio:
This steals the fix from turbo-list
fixes the issue where this applauncher would get stuck when swiping down from the watchface

### DIFF
--- a/src/applauncher/004-three-by-three.qml
+++ b/src/applauncher/004-three-by-three.qml
@@ -62,6 +62,16 @@ Item {
     property bool forbidRight:     false
     */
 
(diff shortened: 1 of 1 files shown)


dodoradio:
yes, sorry I didn't immediately find that issue listing

dodoradio:
this last push trimmed most of the lines that were just copied from turbo-list, as I understood that they weren't actually relevant

beroset (approved):
Tested on catfish and it does indeed improve the behavior.  Thank you!

PR #175, opened 2024-09-08
compositor: Smoothen circle border

MagneFire:
Smoothen borders such that the display blends into the bezel.

This PR is mostly meant as a means to start a discussion on if we want this functionality in the first place :smile:

If we want this we probably also need to think about non-circular displays. Perhaps we can also enable this based on a device property. For instance this feature might make sense on `sawfish` (because of the huge bezel) but not on `smelt` which doesn't have a bezel (flat tire).

### DIFF
--- a/src/qml/compositor/CircleMaskShader.qml
+++ b/src/qml/compositor/CircleMaskShader.qml
@@ -34,11 +34,12 @@ import org.asteroid.controls 1.0
 import org.asteroid.utils 1.0
 
 ShaderEffect {
-    property real smoothness: 0.0007
+    property real smoothness: 0.05
     property bool keepInner: true
+    property real radius: 0.51
 
-    property real end: 0.25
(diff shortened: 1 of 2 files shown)


FlorentRevest:
I originally introduced this for my bass where the screen borders seemed heavily pixelized, this was a cheap trick to make the screen look a bit better than it was :p (this was not about the thickness of the bezels) I still think it adds a nice polish to the AsteroidOS UI but maybe Timo has opinions on this ? At the time, I did not worry about flat tires but it could make sense to add a similar smoothing effect at the level of the flat tire maybe ?

Although it could make sense to compare a few smoothing options, I'd be leaning towards keeping this thing :)

FlorentRevest:
Closing this as stale since the underlying code changed quite a bit already but I agree the conversation around how to smooth the screen should maybe still happen (the use of a shader may have a bigger performance impact than we'd have wanted)

PR #176, opened 2024-09-10
PanelsGrid: Ensure other panels stay hidden in ambient mode

MagneFire:
Ensures that panels other than the current one stay hidden or are hidden whenever in ambient mode. This is required as ambient mode alters the content X and Y, which resulted in showing other components (applauncher) while in ambient mode.


FlorentRevest:
Uhuh, hacks on hacks on hacks :D This PanelsGrid might be the code base I'm the least proud of. This is so complex, hard to reason about and slow. I've always thought that a rewrite would be necessary but I wouldn't even know where to start...

Anyway, thank you for all the fixes. I know this requires painful debugging and code reading :/

PR #180, opened 2025-04-28
Rework QuickSettings and rename to QuickPanel

moWerk:
# Rework QuickSettings.qml for Enhanced Customization and UX in asteroid-launcher

This pull request overhauls `QuickSettings.qml` in the `asteroid-launcher` repository, delivering a redesigned QuickSettings panel for AsteroidOS. While staying fully backwards compatible with no feature regression in function and design over our 10 years old UX many users are familiar with and grew to like. The redesign introduces a flexible, user-customizable interface with a two-row layout, advanced toggle/button interactions, and visual enhancements, all seamlessly integrated with the new `QuickSettingsPage.qml` in `asteroid-settings`. The changes leverage new `qml-asteroid` components (`ValueMeter`, `Particles`, `RemorseTimer`, `OptionsCycler`) to provide a polished, touch-friendly smartwatch UX. Below, we detail the new panel features and the revamped toggle/button components.

## This PR depends on:
https://github.com/AsteroidOS/qml-asteroid/pull/58 to add the necessary components to qml-asteroid.
https://github.com/AsteroidOS/asteroid-settings/pull/104 to provide a settings page for the new user defined order logic.
Already merged:
https://github.com/AsteroidOS/asteroid-icons-ion/pull/23 added a quicksetting logo in two variants to reflect the optional vertical mirrored orientation of the QuickSettings panel layout.
https://github.com/AsteroidOS/asteroid-icons-ion/pull/25 added a Always on Display icon used in the aodToggle.

## Issues adressed:
Resolves https://github.com/AsteroidOS/asteroid-launcher/issues/116 by implementiong all mentioned features while retaining the lockButton feature.
Resolves https://github.com/AsteroidOS/asteroid-settings/issues/45 by adding granular adjustment of the brightnessToggle via PressedAndHold.
Resolves https://github.com/AsteroidOS/asteroid-settings/issues/76 by adding the soundToggle with mute/unmute option via checked/unchecked state additional to the PressAndHold granular volume adjustment.
Resolves https://github.com/AsteroidOS/asteroid/issues/232 by adding a musicToggle, giving users faster access to the asteroid-music application.
Prepares https://github.com/AsteroidOS/asteroid-sounds/issues/2 inclusion of sound files in the stock delivery.
Prepares https://github.com/AsteroidOS/asteroid-settings/issues/32 UI for wifi setup by offering a wifiToggle to conveniently toggle wifi powered state on demand.
[...]

PR #183, opened 2025-07-12
QuickSettings: Increase minimum brightness to 10 as in settings.

### LINKED ISSUE #182: Turning off "sunny" (brightness) quick setting makes screen completely black on lenok
The "sunny" quick setting is hardcoded to 2 values: 100 and 0. While 100% for sunny mode is OK, setting the brightness all the way down to 0% for non-sunny mode makes the screen completely black and the OS unusable (unless you remember where the button was so you can turn it back to sunny and then set a reasonable brightness through the full settings – if you do not, you might even think the OS has crashed altogether) at least on the LG G Watch R (lenok), and probably on some other watches as well.

The full settings actually do not allow setting the brightness below 10% (presumably for this very reason), so at the very least the quick setting should also use 10% instead of 0%. A smarter approach might be to remember the brightness that was set before sunny mode was turned on and switch back to that. But just changing the hardcoded value from 0 to 10 would be a very quick fix.

kkofler:
src/qml/quicksettings/QuickSettings.qml
(rootitem.brightnessToggle.onUnchecked): Set displaySettings.brightness to 10 instead of 0, matching the minimum value in the asteroid-settings app. Setting this to 0 leads to an unusable completely black screen on some hardware, e.g., LG G Watch R (lenok).

Fixes #182.

### DIFF
--- a/src/qml/quicksettings/QuickSettings.qml
+++ b/src/qml/quicksettings/QuickSettings.qml
@@ -83,7 +83,7 @@ Item {
         anchors.verticalCenter: rootitem.verticalCenter
         icon: "ios-sunny"
         onChecked: displaySettings.brightness = 100
-        onUnchecked: displaySettings.brightness = 0
+        onUnchecked: displaySettings.brightness = 10
         Component.onCompleted: updateBrightnessToggle()
     }
 


moWerk:
@kkofler Thanks for pointing out my error! Great catch. As per commit message, my intention was to set all occurrences of minimum brightness to 10. And i missed the main one, nice.

My PR is not ready to merge yet since it needs some more reviews and eventual bugfix commits.
I would ask to correct that error in my own PR and add you as co-author if thats ok with you?

@FlorentRevest we are already working around the 0% brightness issue in the displayPage, but never touched the brightnessToggle for completeness. Afair, some watches have crashing MCE or the light sensor reporting wrong values. Anyway, afair its an issue we decided to work around proactively since the difference between 0 and 10 is barely noticable but the eventual issue of a fully black screen is a major problem.

kkofler:
See PR moWerk#1.

MagneFire:
> @kkofler Thanks for pointing out my error! Great catch. As per commit message, my intention was to set all occurrences of minimum brightness to 10. And i missed the main one, nice.
> 
> My PR is not ready to merge yet since it needs some more reviews and eventual bugfix commits. I would ask to correct that error in my own PR and add you as co-author if thats ok with you?
> 
> @FlorentRevest we are already working around the 0% brightness issue in the displayPage, but never touched the brightnessToggle for completeness. Afair, some watches have crashing MCE or the light sensor reporting wrong values. Anyway, afair its an issue we decided to work around proactively since the difference between 0 and 10 is barely noticable but the eventual issue of a fully black screen is a major problem.

Have to agree with @FlorentRevest on this one. The solution to this issue is to provide a correct mce configuration for the watch that has this issue. Imo we should try to avoid hacks to Qt code that are really hardware related.
For some watches (most notably `sturgeon` and `smelt`) we provide a custom mce configuration that results in correct mapping of 0-100% brightness levels.

Given that these issues have been here for some time now I propose we could do the following:
- Provide a new default mce configuration that just raises the minimum brightness for all watches.
  - Has the advantage that we don't have to test this on all watches. Obvious disadvantage is that the minimum brightness on properly functioning watches has also increased.
- Provide a per watch bases solution
  - Not as trivial as it might seem, more time consuming to test per device (but definitely ideal)
- Solve in Qt/QML code
  - Solves hardware/low-level issues in high-level application code, which is generally not a good idea.

In my opinion the best solution is the combination between the first two: Provide a new mce default and anyone can provide an ideal configuration for their watch.

Another idea would be to apply the Qt/QML solution _for now_. And solve it differently in the future. But the issue with this approach is that it probably will never be fixed.

MagneFire:
https://github.com/AsteroidOS/asteroid-launcher/pull/180 has been merged that should improve the situation.

PR #214, opened 2026-04-02
[Watchfaces performance] watchface code quality and performance refactor

moWerk:
Updated 2026-04-21

## Overview

The `WallClock.enabled` fix that motivated this PR has already merged separately:
binding `enabled: desktop.visible` in MainScreen.qml stops the clock ticking behind
active apps. Verified with dodoradio via console log: `desktop.visible` is `false`
when an app is foregrounded and `true` during peek-to-homescreen and whenever the
launcher surface is active. The redundant `if (!visible) return` guards that
pre-existed in several watchfaces have been removed as part of this PR.

This PR refactors all 18 numbered watchfaces (000–017) for performance, correctness,
and code quality. All license headers are converted from LGPL/BSD prose blocks to
SPDX identifiers.

## SPDX license header conversion

All 18 watchfaces converted from multi-line prose license blocks to compact SPDX
identifiers. Duplicate copyright entries for the same author consolidated to the
oldest year. The 2012 triple-author line expanded to three individual
`SPDX-FileCopyrightText` lines throughout.

## Performance

**Canvas arcs replaced with declarative Shape+PathAngleArc** in
analog-scientific-v2, analog-weather-satellite, and contemporary-digital-rings.
Static full-circle backgrounds and battery/progress arcs are now declarative
`PathAngleArc` bindings. Eliminates `requestPaint` callbacks, stored arc
properties, and JS execution that previously ran every second.

**DropShadow samples reduced to 9** across all watchfaces. Previous values ranged
from 12 to 41. At watch screen resolution 9 samples are indistinguishable from
higher counts and the shader cost scales linearly with sample count.

**`smooth: true` and `antialiasing: true` removed** from all Canvas items where
present. `smooth` has no effect on Canvas elements. `antialiasing` on a Canvas
forces unnecessary compositor work. Both were removed from hourCanvas, minuteCanvas,
secondCanvas, and static stroke canvases throughout.

**`textureSize: Qt.size(width*2, height*2)` removed from all nightstandMode layer
blocks**. Double-resolution offscreen texture upload caused visible stutter on
nightstand mode activation. Removed from all affected watchfaces.

**DropShadow layer removed from continuously animating second hands** in
analog-circle-shades, analog-scientific-v2, and analog-aviator. A layer with a
DropShadow effect on a 16ms-updating Image forces a full 60fps shader recomposite
on every frame. Second hand Images now have no layer.

**16ms sweep Timer added** to analog-circle-shades, analog-scientific-v2, and
analog-weather-satellite, replacing `Behavior on angle`. Uses `new Date()` for
millisecond precision, eliminating the catch-up animation that played on watchface
return after visiting an app.

**Live `wallClock.time` reads moved out of Canvas `onPaint` handlers** throughout.
All Canvas `onPaint` bodies now read only from stored properties (`property int
hour`, `property string ap`, `property string dateText` etc.) that are assigned and
change-gated in the `Connections` block. Prevents redundant clock method calls
during repaint cycles.

**Orbiting text trig bindings moved from property declarations to Connections** in
analog-circle-shades and bold-hour-bebas. `Math.cos/sin` expressions in property
bindings re-evaluated every second; moved to imperative assignments in `Connections`
and `Component.onCompleted`.

**`currentMonth` and `currentDayName` lifted to root scope** in analog-scientific-v2,
eliminating 31 per-second `wallClock.time` reads across all day and month delegate
items.

## Correctness fixes

**`monthBox.month` format string bug fixed** in analog-scientific-v2. `"mm"` (minutes)
was used instead of `"MM"` (month), causing the monthly progress arc to repaint
every minute instead of every month.

**`hourHand.hour = hournightstandMode` typo fixed** in analog-50s-americana. The
concatenated identifier evaluated to `undefined` in JS, which coerced to `0` on a
QML `int` property, leaving the hour hand permanently at midnight. Present since
the watchface was added.

**Identical-branch source ternary fixed** in analog-aviator. Both branches of the
hour and minute SVG source expressions resolved to the ambient variant, meaning the
colored hands never displayed.

**`ctx.shadowColor = (0,0,0,.25)` JS comma-operator bug fixed** in
contemporary-digital-rings. The comma operator discards all but the last operand,
so `shadowColor` was assigned `0.25` — a number — which the Canvas API silently
ignores. Canvas shadows have never worked on this watchface.

**Missing `readonly property var colorArray` added to chargeArc** in
masked-spartan. The `ShapePath.strokeColor` binding referenced `chargeArc.colorArray`
which was never declared, causing a silent crash in nightstand mode.

**Operator precedence bug fixed** in funky-town-words, contemporary-digital-rings,
and analog-words `burnInProtectionManager` bindings. `width * condition ? a : b`
evaluates as `(width * condition) ? a : b` not `width * (condition ? a : b)`.
Parentheses added throughout.

**Beluga full-screen layout fixed** in masked-spartan and contemporary-digital-rings.
The square root Item constraint left 20px strips at top and bottom on Beluga
(320×360). masked-spartan restructured with `anchors.fill: parent` root and a
`Math.min(width, height)` square reference for text geometry.
contemporary-digital-rings arc `PathAngleArc` radius references corrected from
unresolvable `parent.width/height` inside `ShapePath` to the named root Item id.

**AoD compositor glitch fixed in masked-spartan**. The ShaderEffect layer is
disabled when `displayAmbient` is true, rendering plain white numerals on the dark
ambient background instead of running the stencil shader against an absent wallpaper.

**Redundant `if (!visible) return` wallClock guards removed** from all watchfaces
where present. These were made redundant by the `WallClock.enabled: desktop.visible`
fix in MainScreen.qml that has already merged.

## Code quality

All `Connections` blocks updated to Qt 5.15 `function onSignalName()` syntax,
replacing the deprecated Qt 5.14 `onSignalName: { }` slot form which generates
build warnings.

`layer { enabled: true; samples: 4 }` sub-object blocks flattened to
`layer.enabled: true` / `layer.samples: 4` dot-notation throughout. Sub-blocks
are appropriate only when nesting more than two properties.

`property var` changed to `property int` on all `chargecolor` properties across all
segmentedArc Repeaters, and on `hour`, `minute`, `second` stored properties on
Canvas items throughout.

`if`/`else` expressions in property bindings converted to ternary operators across
all watchfaces. Multi-line ternary continuations aligned consistently.

`==` and `!=` replaced with `===` and `!==` strict equality in all `Connections`
handlers and `for` loop conditions throughout.

`Text` items with two-property `anchors {}` blocks converted to inline
`anchors.centerIn` or expanded `anchors { }` per the number of properties.
Property ordering normalised: `anchors`, `visible`, `color`, `opacity`, then
`font {}`, then `text` throughout.

Trailing whitespace removed from all 18 files. Kate editor auto-indent had
introduced whitespace-only lines throughout previous commits.

UTF-8 BOM removed from analog-weather-satellite.

Unused `property real radian` removed from watchfaces where Canvas arcs were
eliminated. Unused Mce imports (`MceBatteryState`, `MceCableState`) removed from
funky-town-words.

`z: 0` default-value declarations removed from numerals-duo-synth-neon-green.
The deliberate `z: 2` on analog-words `circleBack` is retained and documented
with a comment, as it is a legitimate use of z-ordering for dynamic layer
management that declaration order alone cannot satisfy.

Design comment added to masked-spartan noting that the ShaderEffect stencil
requires a non-black wallpaper to be visible, as the numeral shapes are cut from
the wallpaper color.

### DIFF
--- a/src/watchfaces/000-default-digital.qml
+++ b/src/watchfaces/000-default-digital.qml
@@ -1,34 +1,8 @@
-/*
- * Copyright (C) 2023 - Timo Könnecke <github.com/eLtMosen>
- *               2022 - Darrel Griët <dgriet@gmail.com>
- *               2022 - Ed Beroset <github.com/beroset>
- *               2017 - Florent Revest <revestflo@gmail.com>
- * All rights reserved.
- *
- * You may use this file under the terms of BSD license as follows:
- *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions are met:
- *     * Redistributions of source code must retain the above copyright
- *       notice, this list of conditions and the following disclaimer.
- *     * Redistributions in binary form must reproduce the above copyright
- *       notice, this list of conditions and the following disclaimer in the
- *       documentation and/or other materials provided with the distribution.
- *     * Neither the name of the author nor the
- *       names of its contributors may be used to endorse or promote products
- *       derived from this software without specific prior written permission.
- *
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
- * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
- * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
- * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE LIABLE FOR
- * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
- * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
- * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
- * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
- * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
- * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
- */
+// SPDX-FileCopyrightText: 2023 Timo Könnecke <github.com/moWerk>
+// SPDX-FileCopyrightText: 2022 Darrel Griët <dgriet@gmail.com>
+// SPDX-FileCopyrightText: 2022 Ed Beroset <github.com/beroset>
+// SPDX-FileCopyrightText: 2017 Florent Revest <revestflo@gmail.com>
+// SPDX-License-Identifier: BSD-3-Clause
 
 import QtQuick 2.15
 import QtQuick.Shapes 1.15
@@ -41,15 +15,14 @@ Item {
     anchors.fill: parent
 
     function twoDigits(x) {
-        if (x<10) return "0" + x;
-        else      return x;
+        return x < 10 ? "0" + x : "" + x
     }
 
     function prepareContext(ctx) {
         ctx.reset()
         ctx.fillStyle = "white"
         ctx.textAlign = "center"
-        ctx.textBaseline = 'middle';
+        ctx.textBaseline = "middle"
         ctx.shadowColor = "black"
         ctx.shadowOffsetX = 0
         ctx.shadowOffsetY = 0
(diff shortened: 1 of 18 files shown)


src/watchfaces/000-default-digital.qml
```diff
@@ -110,40 +109,37 @@ Item {
 
             Canvas {
                 id: amPmCanvas
-
+
```
MagneFire: Empty lines introduced here. Please get rid of them :D

MagneFire (commented):
Can you split this PR into two PRs?
As one is a trivial fix while the other probable will take more time to get done.

imo https://github.com/AsteroidOS/asteroid-launcher/pull/214/changes/d8bf778cce7a0d03eb4af978c1ebd5c6e5969d5c is an awesome fix we can merge quickly while the others will take more time due to the amount of changes.

src/watchfaces/002-analog-70s-classic.qml
```diff
@@ -33,26 +34,23 @@ import Nemo.Mce 1.0
 
 Item {
     anchors.fill: parent
-
+
```
MagneFire: There's more empty lines, please fix for the entire PR.

FlorentRevest:
Having split this PR by file rather than by change makes this essentially impossible to review methodically uh. 

But anyway, the changes listed in the PR descriptions would all make sense and scrolling through the watchfaces nothing seemed particularly scary. 

Assuming the watchfaces are still functional (I suppose you have tested the PR), I'm ok with merging this as-is.

moWerk:
All changes moved to individual PRs per Watchface which already got merged. Thanks everyone.

PR #215, opened 2026-04-03
[QuickPanel] Document logic and rework buggy long-press granular adjustment

moWerk:
## Replace QuickPanel long-press scrubber with drag-based value control

Replaces the hold-and-auto-increment pattern for brightness and volume
adjustment with a pressAndHold-activated horizontal drag scrubber,
bringing the interaction model in line with the IntSelector component
used elsewhere in the settings UI.

### Interaction model

Holding a brightness or volume toggle activates the scrubber. The full
panel width is the drag range — the fill position and finger position
correspond truthfully regardless of which page or slot the toggle
occupies. A 15% fat-finger margin on each side ensures min and max are
reachable before the screen edge. Releasing ends the scrub cleanly
without firing the toggle's click action.

Swiping the sliding row before the hold threshold fires is unaffected —
the ListView steals the gesture naturally since `preventStealing` is
only set after hold is confirmed.

### Visual transition

On scrub start, all panel elements (toggle rows, ValueMeter, pageDots)
fade out simultaneously. A morphRect pill animates from the ValueMeter's
position and fill color, expanding and darkening into the scrubCover
position. The scrubCover crossfades in over the morphRect as it arrives.
On scrub end the sequence reverses symmetrically, the morphRect shrinking
back to where the ValueMeter reappears.

The scrubCover displays a proportional fill tracking the live value. A
live value label occupies the fixed row area (always visible above the
finger). A mode label at the ValueMeter position confirms which parameter
is being adjusted. Both labels fade with the inverse timing of the panel
elements they replace.

### Behaviour changes

- Scrubbing volume to any value above zero while muted implicitly
  unmutes, clearing `preMuteLevel`. A confirmation sound plays on scrub
  end mirroring the click-to-unmute behaviour.
- Brightness minimum respects the system-enforced floor from
  meta-smartwatch — `rangeMin: 10`, toggling off sets to `rangeMin`.
- Cinema toggle `onClicked` branch ordering corrected following the
  `checked`-based API migration from the dodoradio PR.
- ValueMeter state machine for brightness and volume display removed.
  ValueMeter now shows battery exclusively and fades during scrub.

### Videos

[battery bottom layout]

https://github.com/user-attachments/assets/2e2d0f28-8692-449f-b2aa-a982b922d49e

[battery top layout]

https://github.com/user-attachments/assets/03b516eb-13f1-40e5-9c4b-8cd2ce207c58

### DIFF
--- a/src/qml/quickpanel/QuickPanel.qml
+++ b/src/qml/quickpanel/QuickPanel.qml
@@ -401,11 +401,15 @@ Item {
         valueLowerBound: 0
         valueUpperBound: 100
         anchors.horizontalCenter: parent.horizontalCenter
+        // timer lives inside valueMeter and is exposed as a property so toggle
+        // components can restart it without reaching into panel root scope
         property Timer fadeOutTimer: fadeOutTimer
 
         Timer {
             id: fadeOutTimer
             interval: 2000
+            // state "" returns to default which rebinds value to battery percentage
+            // and re-enables charging animations
             onTriggered: {
                 valueMeter.state = ""
 
@@ -416,7 +420,8 @@ Item {
 
         value: batteryChargePercentage.percent
 
-        // Signal to notify toggles to reset direction
+        // signal broadcast to all range-based toggles to reset their scrub
+        // direction after the fadeOutTimer returns the meter to battery mode
         signal resetDirection
 
         // Animate value changes for smooth fill width transitions
@@ -451,7 +456,12 @@ Item {
         Behavior on fillColor {
             ColorAnimation { duration: 300 }
         }
-
+        
+        // -- ValueMeter state machine --
+        // Replaces the old animate-flag pattern. PropertyChanges inside a State
+        // are withheld by the animation engine during transitions, so value and
+        // text only update after the fade-out step completes. A plain binding
+        // cannot provide this ordering guarantee.
         states: [
             State {
                 name: "brightness"
@@ -480,6 +490,9 @@ Item {
                     PropertyAnimation { target: valueMeterCaption; property: "text"; duration: 0 }
                     PropertyAnimation { target: flashIcon; property: "visible"; duration: 0 }
                 }
+                // duration: 0 is intentional — this step commits the deferred
+                // PropertyChanges (value, text, visibility) between the two fade steps
+                // without consuming any frame time
                 ParallelAnimation {
                     NumberAnimation { target: valueMeter; property: "opacity"; duration: 125; to: 1 }
                     NumberAnimation { target: valueMeterCaption; property: "opacity"; duration: 125; to: 1 }
@@ -564,6 +577,8 @@ Item {
                 }
             }
 
+            // binding keeps rangeValue current so press-and-hold always scrubs
+            // from the real current brightness, not a snapshot from last interaction
             rangeValue: displaySettings.brightness
 
             onPressed: valueMeter.state = "brightness"
@@ -611,8 +626,13 @@ Item {
             icon: wifiStatus.connected ? "ios-wifi" : "ios-wifi-outline"
 
             checkable: true
+            // -- Toggle pattern --
+            // checked binds directly to the system source of truth. No Connections
+            // block needed to stay in sync with external state changes.
             checked: wifiStatus.powered
 
+            // !checked reads the value before the click flips it, so this is
+            // effectively: set powered to whatever it currently isn't
             onClicked: {
                 wifiStatus.powered = !checked
             }
--- a/src/qml/quickpanel/QuickPanel.qml
+++ b/src/qml/quickpanel/QuickPanel.qml
@@ -51,11 +51,117 @@ Item {
     property bool forbidRight: true
     property int toggleSize: Dims.l(28)
 
+    // Set true while a range toggle is scrubbing. Single source of truth that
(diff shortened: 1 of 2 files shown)


FlorentRevest:
I think I like this interaction better yes. The code is already a bit crazy but it doesn't make it much crazier so that looks OK to me

moWerk:
Rebased onto current master (post-ngf/systemsettings migrations, one small semantic conflict resolved in the haptics toggle). Two additions since your review: a Qt6 strict-coercion fix (`| 0` on the imperative rangeValue write — without it scrubbing silently does nothing on Qt6) and a haptic confirmation at hold-detection so the finger knows the scrub is armed before the visual lands. Hardware-tested on skipjack and beluga (2026-07-19): full-width tracking, mute-clear on volume scrub, swipe-cancel cleanup, batteryTop layout all green. Ready for review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

src/qml/quickpanel/QuickPanel.qml
```diff
@@ -403,11 +403,15 @@ Item {
         valueLowerBound: 0
         valueUpperBound: 100
         anchors.horizontalCenter: parent.horizontalCenter
+        // timer lives inside valueMeter and is exposed as a property so toggle
```
FlorentRevest: I find this comment doesn't add any value "timer lives inside valueMeter"... well, yes, it's literally a timer inside valueMeter. This is noisy and drowns signal. The property pattern is also common practice so I don't see the value in explaining it.

src/qml/quickpanel/QuickPanel.qml
```diff
@@ -453,7 +458,12 @@ Item {
         Behavior on fillColor {
             ColorAnimation { duration: 300 }
         }
-
+        
+        // -- ValueMeter state machine --
```
FlorentRevest: This is also useless, it's clear that this is a state machine inside ValueMeter

src/qml/quickpanel/QuickPanelToggle.qml
```diff
@@ -37,6 +37,8 @@ MouseArea {
     height: width
 
     property alias icon: ic.name
+    // checkable defaults false so non-toggle buttons render at full opacity
+    // without the caller needing to force a workaround state
```
FlorentRevest: This is also fairly obvious to me and doesn't deserve a comment

src/qml/quickpanel/QuickPanelToggle.qml
```diff
@@ -37,6 +37,7 @@ MouseArea {
     height: width
 
     property alias icon: ic.name
+
```
FlorentRevest: This commit is littered with formatting changes that drown the actual change

src/qml/quickpanel/QuickPanelToggle.qml
```diff
@@ -56,8 +56,6 @@ MouseArea {
     // (typically rootitem.width) so the full range maps across the panel.
     property bool scrubbing: false
     property int scrubWidth: 0
-    property int startScrubX: 0
-    property int startValue: 0
```
FlorentRevest: Since this code is anyway introduced in the previous commit - it's confusing to have a PR that introduces code and then removes it. This commit should probably have been squashed into its parent

FlorentRevest (commented):
Well, I have some general PR health comments but it's about the form more than about the substance... Honestly this whole QuickPanel codebase is already so incredibly hairy that I can't follow what's being changed. This code is too spaghetti to be reviewed already and it's a lost cause for review :/ it's a bit unfortunate to have such a dark corner in Asteroid but we already gave up when we decided to merge the QuickPanel codebase, this only adds to the mess...
Since it's anyway a lost cause I'm tempted to merge this as is but this is a bit disheartening. I've observed various issues with the quick panel already (inconsistent state handling that probably finds its root in total spaghetti code) and performance issues (like the battery charging animation waking up the CPU many times while it's not even visible, this may actually make battery charging a bit slower...)
In the future we need to avoid merging huge blocks of code like this which become dark corners of our codebases and instead proceed with slow incremental changes that build on good abstractions rather than hundreds of indirections, super complicated semantics and hairy cross-references between all sorts of Items.

src/qml/quickpanel/QuickPanel.qml
moWerk: Both of these comments(valuemeter and state machine) are already gone in the current version.

src/qml/quickpanel/QuickPanel.qml
moWerk: Agreed, but was already removed.

src/qml/quickpanel/QuickPanelToggle.qml
moWerk: Understood, removed.

src/qml/quickpanel/QuickPanelToggle.qml
moWerk: Gone. Squashing should have removed the indent issues. Also removed the spaces.

src/qml/quickpanel/QuickPanelToggle.qml
moWerk: Folded that commit into its parent, thank you.

moWerk:
down to 2 commits, and the diff against what you reviewed is only these cleanups

PR #216, opened 2026-04-03
QuickPanelToggle: handle press-and-hold stealing

dodoradio:
the flickable of the listview steals gestures when it detects a swipe, which doesn't trigger onReleased. Add a handler for onCanceled to fix the issues this causes.


MagneFire (approved):
Awesome, thanks for the quick fix!

PR #217, opened 2026-04-04
asteroid-launcher: stop wallClock ticking when launcher grid is not shown

moWerk:
Bind wallClock.enabled to desktop.visible instead of hardcoded true, clock now pauses when an app is shown and starts again on peek or when launcher gird is visible


MagneFire:
Awesome fix thanks!

PR #218, opened 2026-04-13
Fix quick panel swipe and button interaction

KonTy:
The content Item in PanelsGrid had no explicit size so hit-testing never reached child items (contains() returns false for a 0x0 item). Give it 3x3 panel dimensions and offset panel positions by +1 to keep local coords non-negative.

Release the mouse grab after swipeReleased so clicks arent swallowed by GestureFilterArea. Add preventStealing and pressDelay so toggle buttons actually work inside the flickable.

Also fix the QtGraphicalEffects import to 1.0 which matches the version that ships with the image.

Related PRs:
- https://github.com/AsteroidOS/meta-smartwatch/pull/299 (emulator machine config fix)
- https://github.com/AsteroidOS/asteroid/pull/337 (run-emulator.sh launch script)

### DIFF
--- a/src/gesturefilterarea.cpp
+++ b/src/gesturefilterarea.cpp
@@ -159,6 +159,7 @@ void GestureFilterArea::mouseReleaseEvent(QMouseEvent *event) {
                 currVel = m_velocityY;
             emit swipeReleased(m_horizontal, currVel, m_tracing);
             m_pressed = false;
+            ungrabMouse();
         }
     }
 }
--- a/src/qml/quickpanel/QuickPanel.qml
+++ b/src/qml/quickpanel/QuickPanel.qml
@@ -345,6 +345,7 @@ Item {
         snapMode: ListView.SnapOneItem
         clip: true
         interactive: true
+        pressDelay: 200
         boundsBehavior: Flickable.StopAtBounds
         spacing: Dims.l(4)
 
--- a/src/qml/quickpanel/QuickPanelToggle.qml
+++ b/src/qml/quickpanel/QuickPanelToggle.qml
@@ -35,6 +35,7 @@ MouseArea {
 
     width: parent.width
     height: width
+    preventStealing: true
 
     property alias icon: ic.name
     property bool checkable: false


src/qml/quickpanel/QuickPanel.qml
```diff
@@ -30,7 +30,7 @@
  */
 
 import QtQuick 2.9
-import QtGraphicalEffects 1.15
+import QtGraphicalEffects 1.0
```
FlorentRevest: Ooc - did you actually encounter an issue with this ?

---

Best commit practices would make this an atomic commit so things are easier to review, bisect, revert in the future.

src/gesturefilterarea.cpp
```diff
@@ -96,7 +96,7 @@ void GestureFilterArea::mouseMoveEvent(QMouseEvent *event) {
 
     m_velocityX = (m_velocityX*(m_counter-1) + (event->localPos().x()-m_prevPos.x()))/m_counter;
     m_velocityY = (m_velocityY*(m_counter-1) + (event->localPos().y()-m_prevPos.y()))/m_counter;
-    if(m_tracing) {
+    if(m_tracing && m_counter >= 1) {
```
FlorentRevest: It looks like your description is lacking an explanation for this change

src/qml/quickpanel/QuickPanel.qml
KonTy: Yes — Qt 5.12 (which the image ships) provides QtGraphicalEffects 1.0, not 1.15. Importing 1.15 produces a runtime warning:

\`\`\`
module "QtGraphicalEffects" version 1.15 is not installed
\`\`\`

I've split this into its own atomic commit now. The PR has been restructured into 3 separate atomic commits:

1. **QuickPanel: fix QtGraphicalEffects import version** — 1.15 → 1.0
2. **gesturefilterarea: release mouse grab after swipe ends** — ungrabMouse() after swipeReleased
3. **QuickPanel: fix toggle button interaction inside flickable ListView** — pressDelay + preventStealing

I also dropped the changes that turned out to be unnecessary now that the real root cause was found (PR #300 — abs mouse coordinates): the PanelsGrid.qml content Item changes, the \`m_counter >= 1\` check (was a no-op), and the empty \`else if\` block.

src/gesturefilterarea.cpp
KonTy: You're right, the description was lacking. On closer analysis, the \`m_counter >= 1\` change was actually a no-op — \`mouseMoveEvent\` already increments the counter before calling \`filterMouseEvent\`, so it's always ≥ 1 at that point. I've dropped that change entirely.

The real root cause for the broken mouse input in the emulator was that Qt 5.12's QEvdevMouseHandler defaults to relative mode, treating USB tablet absolute coordinates (0–32767) as pixel deltas. The fix is \`QT_QPA_EVDEV_MOUSE_PARAMETERS=/dev/input/event1:abs\` in PR #300.

This PR now only contains genuinely useful fixes:
- \`ungrabMouse()\` after swipe release (prevents stale grab blocking subsequent touch input)
- \`pressDelay: 200\` on the toggle ListView (lets Qt distinguish flick vs tap)
- \`preventStealing: true\` on toggle MouseAreas (prevents parent Flickable from stealing press events)

src/qml/quickpanel/QuickPanel.qml
KonTy: On second thought, you're right — the build system already handles this via the \`do_configure:prepend()\` sed that downgrades all QML imports for Qt 5.12 compatibility. Changing the source itself was wrong — the upstream should stay at 1.15 as intended. Dropped this commit from the PR.

PR now contains only 2 commits:
1. **gesturefilterarea: release mouse grab after swipe ends** — ungrabMouse() after swipeReleased
2. **QuickPanel: fix toggle button interaction inside flickable ListView** — pressDelay + preventStealing

FlorentRevest: FWIW as we migrate to Qt6 all these import versions will soon be dropped entirely so this is just temporary.
The Sed story sounds like a hack so it will be good to eventually get rid of it but let's make this part of the Qt6 upgrade

src/gesturefilterarea.cpp
```diff
@@ -159,6 +159,7 @@ void GestureFilterArea::mouseReleaseEvent(QMouseEvent *event) {
                 currVel = m_velocityY;
             emit swipeReleased(m_horizontal, currVel, m_tracing);
             m_pressed = false;
+            ungrabMouse();
```
moWerk, line 162: This change needs more context before it can be reviewed.
ungrabMouse() in mouseReleaseEvent is a non-obvious choice. Can you describe the specific bug or sequence of events this is fixing? What happens without it?
This touches gesture handling for the whole launcher so it needs a clear
reproduction case, not just a hunch that it helps.

moWerk (changes requested):
The two QML changes (pressDelay + preventStealing) look reasonable as a
solution to accidental toggle activation during swipes, though I'd suggest
trying pressDelay: 100 first and try to observe a difference. 200ms might be noticeable as tap lag on the toggles. These are worth testing on hardware.

The gesturefilterarea.cpp change needs more context before it can be reviewed.

It should be discussed in a separate PR since the C++ change and the QML changes are solving different problems at different layers.

src/gesturefilterarea.cpp
dodoradio: from https://doc.qt.io/archives/qt-5.15/qquickitem.html#ungrabMouse :
```
Note that this function should only be called when the item wants to stop handling further events. There is no need to call this function after a release or cancel event since no future events will be received in any case. No move or release events will be delivered after this function was called.
```
'there is no need to call this function after a release or cancel' -> this change shouldn't be needed

PR #219, opened 2026-04-13
Fix memory leaks, uninitialized members, and shell escaping

KonTy:
Ran cppcheck and clang-tidy on the C++ sources and fixed the warnings:

- **firstrun**: \`m_mceDbus\` was allocated with \`new QDBusInterface(...)\` without a Qt parent, leaking it. Pass \`this\` as parent. Also accept an optional \`QObject *parent\` in the constructor.

- **gesturefilterarea**: POD members (\`m_horizontal\`, \`m_pressed\`, \`m_tracing\`, \`m_counter\`, \`m_velocityX\`, \`m_velocityY\`, \`m_threshold\`) were left uninitialized. Add in-class default initializers.

- **notificationsnoozer**: notification fields (appName, body, summary, hints) were interpolated directly into a shell command string. Escape \`\\\`, \`"\`, \`$\`, and \`\\`\` to prevent injection.

Also fixed two QML issues found during review:

- **QuickPanel**: power-off and reboot used \`remorseTimer.onTriggered.connect()\` which *accumulated* handlers on every click — tapping power-off three times then letting it expire called PowerOff three times. Replace with a property-based approach.

- **NotificationsPanel**: \`firstNotifView\` was reassigned via \`createObject()\` without destroying the previous instance, leaking the old view.

### DIFF
--- a/src/firstrun.cpp
+++ b/src/firstrun.cpp
@@ -33,9 +33,9 @@
 #include <QDBusPendingCall>
 #include <QDBusVariant>
 
-FirstRun::FirstRun()
+FirstRun::FirstRun(QObject *parent) : QObject(parent)
 {
-    m_mceDbus = new QDBusInterface("com.nokia.mce", "/com/nokia/mce/request", "com.nokia.mce.request", QDBusConnection::systemBus());
+    m_mceDbus = new QDBusInterface("com.nokia.mce", "/com/nokia/mce/request", "com.nokia.mce.request", QDBusConnection::systemBus(), this);
     firstRunDone = new MGConfItem("/org/asteroidos/launcher/firstRunDone", this);
 }
 
--- a/src/firstrun.h
+++ b/src/firstrun.h
@@ -38,7 +38,7 @@ class FirstRun : public QObject
 {
     Q_OBJECT
 public:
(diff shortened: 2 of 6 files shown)


FlorentRevest:
This PR seems to contain https://github.com/AsteroidOS/asteroid-launcher/pull/218/changes by mistake.

A better commit hygiene would split `Fix memory leaks, uninitialized members, and shell escaping` into atomic commits.

But otherwise all the fixes themselves look fine, thanks! The notification snoozer shell injection is particularly awkard ugh.

KonTy:
Fixed both issues:

1. Removed the PR #218 content that was accidentally included (the first commit was a duplicate). This PR now only has the static analysis and code review fixes.

2. Split into 5 atomic commits, one per fix:
   - firstrun: fix QDBusInterface memory leak
   - gesturefilterarea: initialize POD members
   - notificationsnoozer: escape shell metacharacters in notification fields
   - QuickPanel: fix remorse timer accumulating handlers on repeated clicks
   - NotificationsPanel: destroy previous view before creating a new one

Force-pushed.

FlorentRevest:
Thanks!

PR #221, opened 2026-04-15
QuickPanel: add airplane mode toggle

KonTy:
Adds an airplane mode toggle to the quick panel that kills all radios with one tap.

It uses ConnMan's offlineMode to turn off WiFi and Bluetooth together. When you flip it back off, it remembers what was on before and restores it. If you manually toggle WiFi or BT while in airplane mode, it won't undo that when you exit basically  it only restores the ones you didn't touch. This is exact behavior that Android and iOS use so it is very intuitive to Android and iOS users

State is saved with ConfigurationValue so it sticks across restarts.

Tested on the QEMU emulator — toggle shows up in the slider row, activates/deactivates cleanly, and the restore logic works as expected.

Please help me test this on some watches. I'm still trying to get a USB dongle for TicWatch 3 hoefully by tomorrow I will be able to start testing my changes on the watch as well. 

Thank you for your help!!!

### DIFF
--- a/src/qml/quickpanel/QuickPanel.qml
+++ b/src/qml/quickpanel/QuickPanel.qml
@@ -30,7 +30,7 @@
  */
 
 import QtQuick 2.9
-import QtGraphicalEffects 1.15
+import QtGraphicalEffects 1.0
 import QtMultimedia 5.8
 import org.asteroid.controls 1.0
 import org.asteroid.utils 1.0
@@ -86,7 +86,7 @@ Item {
     ConfigurationValue {
         id: sliderToggles
         key: "/desktop/asteroid/quickpanel/slider"
-        defaultValue: ["brightnessToggle", "bluetoothToggle", "hapticsToggle", "wifiToggle", "soundToggle", "cinemaToggle", "aodToggle", "powerOffToggle", "rebootToggle", "musicButton", "flashlightButton"]
+        defaultValue: ["brightnessToggle", "bluetoothToggle", "hapticsToggle", "wifiToggle", "soundToggle", "airplaneModeToggle", "aodToggle", "powerOffToggle", "rebootToggle", "cinemaToggle", "musicButton", "flashlightButton"]
     }
 
     ConfigurationValue {
@@ -100,10 +100,11 @@ Item {
             "hapticsToggle": true,
             "wifiToggle": true,
             "soundToggle": true,
-            "cinemaToggle": true,
+            "airplaneModeToggle": true,
             "aodToggle": true,
             "powerOffToggle": true,
             "rebootToggle": true,
+            "cinemaToggle": false,
             "musicButton": false,
             "flashlightButton": false
         }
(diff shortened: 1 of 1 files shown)


moWerk (commented):
I will review in the night, but if you read this earlier -> I would suggest to add your new toggle at the place the Cinema Mode currently is sorted. And then move the cinema mode down (above the flashlight & music short cuts).
And then disable the cinema mode as default. Effectively swapping yours with the cinema mode.
Your airplane mode is of much greater use than the cinema mode and i would have loved to have more toggles disabled for the user to discover and only activate when needed. But enabled the cinema mode just to fill the sliding row ;)

KonTy:
> I will review in the night, but if you read this earlier -> I would suggest to add your new toggle at the place the Cinema Mode currently is sorted. And then move the cinema mode down (above the flashlight & music short cuts). And then disable the cinema mode as default. Effectively swapping yours with the cinema mode. Your airplane mode is of much greater use than the cinema mode and i would have loved to have more toggles disabled for the user to discover and only activate when needed. But enabled the cinema mode just to fill the sliding row ;)

hey thanks for the quick feedback! done: swapped airplane mode into cinema's spot, moved cinema down by the shortcuts, and disabled it by default. airplane mode is enabled by default now.

totally agree it's way more useful than cinema mode in that prime real estate. pushed the update, let me know if it looks good or if you want anything else tweaked

src/qml/quickpanel/QuickPanel.qml
```diff
@@ -86,7 +86,7 @@ Item {
     ConfigurationValue {
         id: sliderToggles
         key: "/desktop/asteroid/quickpanel/slider"
-        defaultValue: ["brightnessToggle", "bluetoothToggle", "hapticsToggle", "wifiToggle", "soundToggle", "cinemaToggle", "aodToggle", "powerOffToggle", "rebootToggle", "musicButton", "flashlightButton"]
+        defaultValue: ["brightnessToggle", "bluetoothToggle", "hapticsToggle", "wifiToggle", "soundToggle", "airplaneModeToggle", "aodToggle", "powerOffToggle", "rebootToggle", "cinemaToggle", "musicButton", "flashlightButton"]
```
moWerk, line 89: Thanks for moving the toggles. This is good and in correct order.

src/qml/quickpanel/QuickPanel.qml
```diff
@@ -100,8 +100,9 @@ Item {
             "hapticsToggle": true,
             "wifiToggle": true,
             "soundToggle": true,
-            "cinemaToggle": true,
+            "cinemaToggle": false,
             "aodToggle": true,
+            "airplaneModeToggle": true,
```
moWerk, line 103: Here i would have expected the same sorting in code order like in the defaultValue. cinemaToggle to the bottom so all false toggles collect there. and new airplane into place of cinemaToggle

src/qml/quickpanel/QuickPanel.qml
```diff
@@ -262,7 +296,8 @@ Item {
             "powerOffToggle": powerOffToggleComponent,
             "rebootToggle": rebootToggleComponent,
             "musicButton": musicButtonComponent,
-            "flashlightButton": flashlightButtonComponent
+            "flashlightButton": flashlightButtonComponent,
```
moWerk: Same here, this order should reflect the visual sorting like defaultValue already does.

moWerk (changes requested):
The approach is right and I'd like to see this merged, but a few things need addressing first.

The four ConfigurationValue objects should be plain properties instead. This state is transient. It only matters while airplane mode is active and should not survive a reboot. Using dconf keys for runtime-only state pollutes the config namespace unnecessarily.

The restore logic (setting wifiStatus.powered and btTechnology.powered after disabling offlineMode) needs hardware testing specifically. ConnMan may restore technology states itself when offlineMode goes false, in which case those calls either conflict or race. QEMU cannot reproduce this. Please test on a real device before we merge.

Also worth checking: does setting wifiStatus.powered = true while offlineMode is true actually do anything? If ConnMan ignores it, the user-changed detection logic in the wifi and bluetooth onClicked handlers is silently broken.

The sliderToggles order is correct. But the placement is inconsistent across the file — airplaneModeToggle needs to sit in cinema's old position in the toggleEnabled map, the toggleRegistry map, and the Component definitions at the bottom of the file too, not just in the sliderToggles array. Code order should mirror UI order throughout so the file stays readable.

Toggle placement and default value changes look good otherwise.

PR #229, opened 2026-05-16
Fix Qt6 crash and type error in 003-turbo-list.qml

moWerk:
Three related Qt6 regressions in the turbo list applauncher.

currentPos is declared as int but onContentYChanged assigned the result of Math.round() which returns a JavaScript Number (double). Qt6 rejects the implicit coercion that Qt5 accepted silently. Fix: append | 0 to force integer truncation before assignment.

During scroll, items that move above the visible center produce a positive itemLocationY causing the bezelOffset discriminant to go negative. Math.sqrt of a negative number returns NaN, which propagates into font.pixelSize and crashes the QSGRenderThread with a null pointer dereference at offset 0x1c. Qt5 text rendering tolerated NaN silently. Fix: guard bezelOffset against a negative discriminant, returning screenRadius (normalizedBezelOffset = 0) for out-of-range items.

font.pixelSize evaluates to zero when normalizedBezelOffset is 0, which is also invalid for text rendering. Fix: wrap in Math.max(1, ...) to ensure a minimum valid size.

### DIFF
--- a/src/applauncher/003-turbo-list.qml
+++ b/src/applauncher/003-turbo-list.qml
@@ -91,7 +91,10 @@ ListView {
         // Finally we add a small padding (Dims.w(5)) so that the item is not touching the left 'bezel'.
         property var screenRadius: appsView.height/2
         property var itemLocationY: (launcherItem.height * (appsView.contentY/launcherItem.height - index) - launcherItem.height/2)
-        property var bezelOffset: screenRadius - Math.sqrt(Math.pow(screenRadius, 2) - Math.pow((screenRadius + itemLocationY),2))
+        property var bezelOffset: {
+            var d = Math.pow(screenRadius, 2) - Math.pow(screenRadius + itemLocationY, 2)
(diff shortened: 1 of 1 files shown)


MagneFire:
Crash is already fixed here: https://github.com/AsteroidOS/asteroid-launcher/pull/227

moWerk:
> Crash is already fixed here: #227

Yes, you ninjaed that while i was fixing, bwahaha.

I resolved the conflict. Are the two remainig changes still welcome?

MagneFire:
> > Crash is already fixed here: #227
> 
> Yes, you ninjaed that while i was fixing, bwahaha.
> 
> I resolved the conflict. Are the two remainig changes still welcome?

If they fix something then yes. But I haven't observed issues from the other PR. 

So, please check if your changes are still needed.

MagneFire:
For instance I suspect that the double to int issues are a red herring. And that it only fails to implicity cast to int when the double values are NaN. But please verify this stuff ;)

moWerk:
Closed since the actual effects could not be observed anymore after  #227

Issue #198, opened 2026-02-06
Screen stays on when putting watch to sleep while an app is open


Christoph-Harms:
### Observed behavior
When putting the watch to sleep (either by Palm-To-Sleep or by pressing the hardware button) while being in the settings menu *and* the watch being on the charger, the screen stays on while the watch goes to sleep. The watch does go to some sort of sleep though: The screen dims slightly and becomes unresponsive to input (except the tap-to-wake) until woken.

This only happens when the watch is in the settings menu *and* charging, removing either of these two conditions leads to the watch going to sleep normally.

### Expected behavior
Watch should go to sleep normally, with the screen turned off

### Further Info
OPPO Watch (beluga)
AsteroidOS 1.1-nightly

moWerk:
> I'd expect that the screen turning off would immediately show the watchface AoD.

> And that turning the screen back on within 5 seconds would show the app again.

Agreed. Pressing the power button is a clear user signal to set the watch to sleep and see the AoD as fast as possible.

PR #235, opened 2026-05-18
Switch to home screen immediately when entering AOD/ambient mode (fixes meta-asteroid#228)

NickASanOS:
When AOD (Always-On Display) activates while an application is open, the
compositor previously kept showing the dimmed application for 5 seconds
until `delayTimer` fired, killed the app, and switched to the home screen.

During those 5 seconds the dimmed app was visible instead of the
AOD-optimized watchface. This was reported as
[AsteroidOS/meta-asteroid#228](https://github.com/AsteroidOS/meta-asteroid/issues/228).

## Changes

- Added `onDisplayAmbientEntered` handler: immediately switches to the home
  window when ambient mode activates, saving a reference to the app
- Modified `delayTimer`: kills the saved app reference instead of the
  current topmost window (which is now the home window)
- Updated `onDisplayAboutToBeOn`: restores the saved app if the display
  wakes before the 5-second cleanup timer fires (but not during periodic
  ambient minute updates)
- Added `onWindowRemoved` safety check: clears the saved app reference if
  the app is closed externally while in ambient mode

Fixes AsteroidOS/meta-asteroid#228

### DIFF
--- a/src/qml/compositor/compositor.qml
+++ b/src/qml/compositor/compositor.qml
@@ -153,7 +153,10 @@ Item {
         interval: 5000
         repeat: false
         onTriggered: {
-            Lipstick.compositor.closeClientForWindowId(comp.topmostWindow.window.windowId)
+            if (comp.ambientAppWindow) {
+                Lipstick.compositor.closeClientForWindowId(comp.ambientAppWindow.window.windowId)
+                comp.ambientAppWindow = null
+            }
             Lipstick.compositor.setAmbientUpdatesEnabled(true)
         }
     }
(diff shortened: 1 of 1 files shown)


MagneFire:
Thank you for working on this!

This doesn't seem to work correctly.

When the display transitions to AOD mode the app is now correctly hidden, but upon display awake the app does not appear to be restored.
My guess is that this occurs because `!Lipstick.compositor.displayAmbient` doesn't evaluate to true.
There's also another issue where the `onDisplayAboutToBeOn` doesn't get fired if you quickly tap the power button.

Another question, how does this PR relate to https://github.com/AsteroidOS/asteroid-launcher/pull/236/changes and https://github.com/AsteroidOS/lipstick/pull/24/changes? They both appear to attempt to solve the same issue (meta-asteroid#228 and #198 refer to the same issue)

PR #257, opened 2026-06-06
Refactor 016-masked-spartan watchface for Qt6

moWerk:
Qt6 refactor of the 016-masked-spartan watchface. Five commits covering SPDX header, Qt6 ShaderEffect fix with build-time shader compilation, nightstand arc correctness and conventions, non-square screen geometry fix, and a qmlformat pass. This is the most complex watchface in the stock set from a Qt6 migration perspective, requiring porting of an inline GLSL shader to the Qt6 shader pipeline.

> **Note:** This PR depends on the `shaders/` folder PR which establishes `src/shaders/` and its CMake install target. Please merge that PR first.

### Commits

1. **Replace legacy copyright block with SPDX header** — converts the multi-line LGPL block comment to `SPDX-FileCopyrightText` and `SPDX-License-Identifier: LGPL-2.1-or-later` one-liners. Eight authors preserved newest-first. Updates Timo Könnecke handle from `eLtMosen` to `moWerk`, creation year 2023 preserved.

2. **Fix Qt6 ShaderEffect crash and add build-time shader compilation** — Qt6 no longer accepts inline GLSL strings in `ShaderEffect.fragmentShader`. The inline GLSL 100 shader was rewritten in GLSL 440 and added as `src/shaders/masked-spartan.frag`. Added to the existing `qt_add_shaders` block in `src/CMakeLists.txt` without a `PREFIX` so the existing `circlemaskshader.frag` resource path is not disturbed. The shader is compiled into the resource bundle at build time and referenced as `qrc:/shaders/masked-spartan.frag.qsb` in the QML. The shader logic: in normal mode (`keepInner: false`) letter shapes are cut out of the dark rectangle revealing the wallpaper behind; in AoD `layer.enabled: !displayAmbient` bypasses the shader entirely, allowing Text items to render directly on the dark background at reduced opacity. `keepInner` is a constant `false` since AoD no longer routes through the shader.

3. **Qt6 nightstand arc correctness and conventions** — removes the `layer` block from `nightstandMode`. Removes dead `batteryPercentChanged` property. Adds `| 0` bitwise cast to `chargecolor` and changes `property var` to `property int`. Removes `smooth: true` and `antialiasing: true` from `chargeArc` Shape. Normalises `colorArray` bracket spacing and `startY` parenthesis spacing. Sorts imports alphabetically. Removes `org.asteroid.controls` which has no references in the file. Retains `org.asteroid.utils` for `DeviceSpecs.hasRoundScreen` used in the `layer2mask` radius binding. Replaces root square sizing ternary with `Math.min`.

4. **Fix non-square screen geometry** — adds `id: faceRoot` to the screen-filling outer Item. Moves `layer2mask` and `_mask` from the inner square Item to become direct children of `faceRoot`, anchored to `parent` which correctly resolves to the full screen surface. On rectangular screens the dark background rectangle now fills the full panel instead of being clipped to the square inner Item. The inner square Item is retained and now contains only `nightstandMode` and `MceBatteryLevel`. Sets `layer.enabled: !displayAmbient` on `_mask` to bypass the shader in AoD. Sets `opacity: displayAmbient ? 0.75 : 1` on `_mask` for correct AoD glyph visibility. `layer2mask` color is now a static `Qt.rgba(0, 0, 0, 0.75)` with no `displayAmbient` switching needed. Removes dead `imgPath` property which had no references in the file.

5. **qmlformat pass** — full `qmlformat -i` normalisation pass over the file.

### Testing

Built on 2.2 nightly Qt 6.11 (sturgeon, 400px), compared against 1.1 nightly Qt 5.15 vanilla (tunny, 400px).

- Normal mode: letter cutout effect renders correctly — wallpaper visible through glyph shapes, dark rectangle fills the screen.
- Hour and minute glyphs, seconds, am/pm, date, and month all render at correct positions.
- 12h/24h toggle: hour glyph and am/pm visibility respond correctly.
- AoD: shader bypassed via `layer.enabled: !displayAmbient`; glyphs render directly as white text at reduced opacity on the dark background. No shader artefacts.
- Nightstand mode: battery arc renders correctly. No multisample renderbuffer journal warning.
- No regressions found across all five commits.

### Notes

- The outer screen-filling `Item` (`faceRoot`) is the system-injected root surface and intentionally fills the full panel. `layer2mask` and `_mask` fill `faceRoot` so the dark background covers the full screen including non-square panels. The inner square `Item` constrains only the nightstand arc geometry.
- `masked-spartan.frag` is the GLSL 440 shader source residing in `src/shaders/`. It is compiled into the resource bundle at build time via `qt_add_shaders`. In Qt6, `ShaderEffect.fragmentShader` requires a `qrc:` resource URL pointing to a compiled `.qsb`; inline GLSL strings are no longer supported.
- The `src/shaders/` folder established in the prerequisite PR is the designated location for shader source files. Once that PR merges, the `.gitkeep` placeholder can be removed as the folder will contain `masked-spartan.frag`.
- This shader compilation workflow via `qt_add_shaders` is the established pattern for any future Qt6 shader work in asteroid-launcher, including the planned FragCoord live wallpaper component.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -18,7 +18,9 @@ add_executable(asteroid-launcher ${SRC} ${HEADERS} resources-qml.qrc)
 
 qt_add_shaders(
     asteroid-launcher "asteroid-launcher_shaders"
-    FILES "qml/compositor/circlemaskshader.frag"
+    FILES
+        "qml/compositor/circlemaskshader.frag"
+        "shaders/masked-spartan.frag"
 )
 
 target_link_libraries(
--- a/src/shaders/masked-spartan.frag
+++ b/src/shaders/masked-spartan.frag
@@ -1,0 +1,21 @@
+#version 440
+
+layout(location = 0) in vec2 qt_TexCoord0;
+layout(location = 0) out vec4 fragColor;
+
+layout(std140, binding = 0) uniform buf {
+    mat4 qt_Matrix;
+    float qt_Opacity;
+    bool keepInner;
+};
+
+layout(binding = 1) uniform sampler2D source;
+layout(binding = 2) uniform sampler2D maskSource;
+
+void main(void) {
+    if (keepInner) {
+        fragColor = texture(source, qt_TexCoord0.st) * (texture(maskSource, qt_TexCoord0.st).a) * qt_Opacity;
+    } else {
+        fragColor = texture(source, qt_TexCoord0.st) * (1.0 - texture(maskSource, qt_TexCoord0.st).a) * qt_Opacity;
+    }
+}
--- a/src/watchfaces/016-masked-spartan.qml
+++ b/src/watchfaces/016-masked-spartan.qml
@@ -19,156 +19,155 @@ import org.asteroid.utils
 
 
 Item {
-    anchors.fill: parent
+    id: faceRoot
 
-    property string imgPath: "../watchfaces-img/funky-town-words-"
+    anchors.fill: parent
+    
+    Rectangle {
+        id: layer2mask
+
+        anchors.fill: parent
+        color: Qt.rgba(0, 0, 0, 0.75)
+        visible: true
+        opacity: 0
+        layer.enabled: true
+        layer.smooth: true
+        radius: DeviceSpecs.hasRoundScreen || nightstand ? faceRoot.width : 0
+    }
+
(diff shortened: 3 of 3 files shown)


moWerk:
Drafted to implement the move of the shader blob to a new general shaders folder

PR #261, opened 2026-06-15
Import Lipstick sources and aggressively simplify the interface it offers

FlorentRevest:


lipstick/CMakeLists.txt
```diff
@@ -0,0 +1,110 @@
+pkg_check_modules(mce REQUIRED IMPORTED_TARGET mce)
```
PureTryOut: Perhaps move these to the root CMakeLists.txt, that is where packagers (like me) would look for dependencies rather than navigating through every subfolder. You can still use the imported variables from them just fine in subfolders.

PureTryOut (changes requested):

lipstick/CMakeLists.txt
FlorentRevest: Done - thanks!

PR #267, opened 2026-06-18
fix(cmake): actually link to the imported ngf-qt6 target

PureTryOut:
This was done differently because it seems AsteroidOS installs the ngf-qt6 headers in a different place than it's .pc file says. That is really a packaging bug and should be fixed in the ngf-qt6 packaging, not worked around here. Without this change it fails to build on platforms where the ngf-qt6 packaging is correct like postmarketOS.

Follow-up to #266


FlorentRevest:
Works with https://github.com/AsteroidOS/meta-asteroid/pull/275 now

PR #270, opened 2026-06-28
Play UI sounds through ngf instead of QtMultimedia SoundEffect

stianlaa:
Hi folks, first PR to AsteroidOS repos, so pointers are very welcome!

Came across this when improving battery lifetime for the `eos` variant of the upcoming `aurora` port. 

The notification and un-mute sounds are played with a QtMultimedia SoundEffect. That object opens a PulseAudio stream when it is created and keeps it open for the launcher's whole lifetime. Because these SoundEffect objects
live for the entire session, their streams never close, the audio sink never suspends

This depends on [AsteroidOS/meta-asteroid#279](https://github.com/AsteroidOS/meta-asteroid/pull/279), which adds the sound-only "notification" ngf event this calls. Without that event the launcher would be silent, so that one should land first.

### DIFF
--- a/src/qml/quickpanel/QuickPanel.qml
+++ b/src/qml/quickpanel/QuickPanel.qml
@@ -31,7 +31,6 @@
 
 import QtQuick
 import Qt5Compat.GraphicalEffects
-import QtMultimedia
 import org.asteroid.controls
 import org.asteroid.utils
 import org.asteroid.launcher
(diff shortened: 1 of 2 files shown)


src/qml/quickpanel/QuickPanel.qml
```diff
@@ -144,10 +143,14 @@ Item {
...
-    SoundEffect {
-        id: unmuteSound
-        source: "file:///usr/share/sounds/notification.wav"
-        volume: 0.8
+    // Play the un-mute confirmation through ngf rather than a QtMultimedia
+    // SoundEffect: a SoundEffect holds a PulseAudio stream open for its whole
+    // lifetime, preventing the audio sink from suspending (constant battery
+    // drain). ngfd plays via a short-lived GStreamer pipeline so the sink can
+    // suspend again at idle. Reuses the "notification" ngf event.
+    NonGraphicalFeedback {
```
stianlaa, line 146: I'm definitively open to a more Qt idiomatic way to shorten this lifetime. First couple of PRs coming up here, so I'd appreciate pointers and feedback. 

In short what I'm after here is to fire and forget this sound effect, and not have it linger, keeping battery drain.

stianlaa:
Selfreviewed, I'll instead try to handle this in Qt fully.

stianlaa:
Reopened as discussed in #271. Rebased onto current master, and the code comments and commit message now spell out both problems this addresses: the Qt6 QSoundEffect chop (AsteroidOS/asteroid#342) and the never-suspending audio sink / standby drain.

Depends on AsteroidOS/meta-asteroid#279, which is reopened as well and should land first.

I'll retest on my Pixel Watch 2 (aurora) and report back here - and happy to take you up on the beluga test offer, @moWerk. @FlorentRevest re asteroid-settings and asteroid-alarmclock: agreed they'd benefit from the same treatment

FlorentRevest (approved):

FlorentRevest:
I took the liberty to minimize your diff but squashed my changes so it's all edited in one commit by you

PR #272, opened 2026-07-04
GestureFilterArea: handle touch natively to fix panel swipes on Qt6

### LINKED ISSUE #237: Qt6: Today Panel can not be left
After lengthy debug log tracing, i found the listView in the today panel seems to steal the swipe from the gesture filter area.
All other flickables -> notification -> quickpanel -> launchers, seems to be awrapped in an item. So maybe wrapping the today view listview in to an item solves the issue.

moWerk:
## Problem

Since the Qt6 migration, panels containing a touch-native item could not be swiped back to the homescreen (#237). The Today panel was the always-broken case, since its ListView accepts touch natively in Qt6 and becomes the exclusive grabber of the touch point.

## Root cause

GestureFilterArea filtered child mouse events only. Under Qt6 it still received temporarily synthesized mouse events while filtering, but when it stole the grab with the deprecated grabMouse(), qtdeclarative cancelled the touch-to-mouse synthesis. As the item did not accept touch events, nothing could reach it anymore and the swipe froze on the first threshold-crossing move. Verified against the exact qtdeclarative revision shipped in the nightlies.

## Fix

Adopt the modern pattern from QQuickFlickable::childMouseEventFilter. GestureFilterArea now accepts touch events, handles TouchBegin/Update/End directly in the filter, takes the grab by accepting the filtered touch event (which lets the delivery agent transfer the grab while respecting a child's keepTouchGrab), and receives post-grab updates natively via a touchEvent() override. Thresholds, velocity math and the swipeMoved/swipeReleased contract are unchanged, and the mouse path remains for pointer devices.

## Validation

A desktop reproducer compiles this file against Qt 6.11 and injects touch through the normal delivery agent. Pre-fix, the Today-analog ListView scenario is fully dead. Post-fix all 17 scenarios pass: swipes work over ListView and MouseArea panels, taps still reach child MouseAreas, and list scrolling in forbidden directions is unaffected. Also built and tested on watch hardware, see below.

Fixes #237

---

Solution found using claude code after pre investigation by moWerk and Dodoradio.
Code changes pre-reviewed, built and tested on watch hardware:
Dodoradio on medaka
moWerk on catfish, beluga and nemo


src/gesturefilterarea.cpp
```diff
@@ -27,6 +27,21 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+/*
+ * Qt6 note: a Flickable/ListView child accepts touch events natively, so the
```
MagneFire: Can you remove this comment. The git commit message already provides this information, and that's enough

src/gesturefilterarea.cpp
```diff
@@ -27,6 +27,21 @@
...
+ * with the deprecated grabMouse(), the delivery agent cancelled the temporary
+ * touch->mouse synthesis (sendFilteredPointerEventImpl ->
+ * cancelTouchMouseSynthesis) and, since GestureFilterArea did not accept
+ * touch events, no further updates could ever be delivered to it: the swipe
+ * froze on the first threshold-crossing move (issue #237).
+ *
+ * The modern pattern (same as QQuickFlickable::childMouseEventFilter) is to
+ * handle touch events directly in the filter, take the grab by accepting the
+ * filtered touch event, and keep receiving updates natively in touchEvent().
+ */
```
FlorentRevest, line 36: I don't think there's a need for this comment. Referring to the code how it used to be doesn't help anyone understand the new code because the old code will be gone the moment we hit merge :p similarly, I don't think we should have a "note for Qt6" when the code base only supports Qt6. Let's just drop this, the rest looks alright

Issue #44, opened 2020-08-06
Screen recording


ChandlerSwift:
More than once, I've had an issue to debug or wanted to post a screen recording demo of what's happening on the watch. AsteroidOS does have [screen*shot* support via D-Bus](https://github.com/AsteroidOS/lipstick/blob/master/src/screenshotservice.cpp), but as far as I can tell no support for screen recordings.

Android has similar functionality that perhaps we could model in the [screenrecord tool](https://developer.android.com/studio/command-line/adb#screenrecord) (which in fact does come with a default build of asteroidos, but doesn't know how to talk to our compositor!)

```
root@harmony:~# which screenrecord
/system/bin/screenrecord
root@harmony:~# screenrecord test.mp4
ERROR: unable to get display characteristics
root@harmony:~#
```

MagneFire:
Recently, I had a look into this, because I needed it for something too :wink: 
Only did some research, but it basically already exists for SailfishOS/Mer Project.

`Lipstick` is base for `asteroid-launcher`. Upstream `Lipstick` has support for screen recording, but this has been removed by @FlorentRevest in this commit: https://github.com/AsteroidOS/lipstick/commit/0274da621e9a26b1386c624fc841fcae89a3f3ac.
You likely only have to re-implement this, here is the upstream variant: https://git.sailfishos.org/mer-core/lipstick/commits/master/src/compositor/lipstickrecorder.cpp

Once that is done, the compositor should have support for screen recording, then all you have to do is create a screen recorder app. Which also exists: https://github.com/nemomobile/lipstick-recorder or https://github.com/CODeRUS/screenrecorder

You will likely need to change the app so that it works in a client/server configuration, as on AsteroidOS we don't support background apps(i.e. screen goes off -> app gets killed).

PR #275, opened 2026-07-05
BluetoothAgent: Limit content to a square so the pairing circle stays round

### LINKED ISSUE #274: bluetooth agent: pairing screen is oval on beluga
https://github.com/AsteroidOS/asteroid-launcher/blob/master/src/qml/misc/BluetoothAgent.qml

needs the same fix as 
https://github.com/AsteroidOS/qml-asteroid/issues/96

moWerk:
On beluga the pairing circle was stretched into an oval and the dialog sat offset to the top. This wraps the content in a centered Math.min(width, height) square and sizes it against that square instead of the former Dims, so the circle renders round again and the dialog centers.

It drops Dims (and the initialSize screen reference) from this component, which is the direction Dodo outlined in #44: Dims assumes fullscreen and breaks as soon as a component is sized any other way (e.g. the watchface/settings preview).

Tested on beluga.

Fixes: https://github.com/AsteroidOS/asteroid-launcher/issues/274

### DIFF
--- a/src/qml/misc/BluetoothAgent.qml
+++ b/src/qml/misc/BluetoothAgent.qml
@@ -35,110 +35,119 @@ import org.nemomobile.lipstick
 Item {
     id: btAgent
 
-    width: initialSize.width
-    height: width
+    anchors.fill: parent
     visible: bluetoothAgent.windowVisible
 
+    // The dialog is sized against a centered square (the shorter screen edge)
+    // instead of Dims, so the pairing circle stays a circle and everything
+    // centers on non-square screens.
     Item {
-        id: circleWrapper
-        anchors.fill: parent
-        Rectangle {
-            id: circle
-            anchors.centerIn: parent
-            width: parent.width*0.7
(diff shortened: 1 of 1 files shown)


src/qml/misc/BluetoothAgent.qml
```diff
@@ -35,110 +35,119 @@ import org.nemomobile.lipstick
...
-        cached: true
-    }
-
-    Icon {
-        id: icon
+        id: content
         anchors.centerIn: parent
-        anchors.verticalCenterOffset: -Dims.h(20)
-        width: Dims.w(20)
+        width: Math.min(parent.width, parent.height)
```
FlorentRevest, line 47: I suppose the flat tire is also not taken into account here?

moWerk: No, that is what irritated me with the RemorseTimer. This element is not full screen and nested in the launcher so the flatTyre offset is already taken account for.

PR #277, opened 2026-07-20
compositor: Gate per-window round-corner mask on motion to kill 2x composite echo

MagneFire:
`WindowWrapperBase` enabled a `CircleMask` `layer.effect` (`smoothBorders`) on every app window and never cleared it. `QQuickItemLayer`'s internal live `QQuickShaderEffectSource` then re-grabbed the app's Wayland surface after every client commit, scheduling a second compositor frame per commit (the "2x composite echo"): with any app foregrounded the compositor rendered roughly twice per client frame.

Bind `smoothBorders` to the window's motion state (opening, closing/minimizing, or being dragged) instead of leaving the layer permanently enabled. At rest the layer is off and the fullscreen app corners are already clipped by the top-level CircleMask; the per-window mask is re-applied only while the window is in motion, so its corners stay round through those transitions.

The compositor is back to one composite per client commit. On minnow, where the DSI link saturates at 45Hz, the extra composites were capping app frame delivery at half rate: a scrolling settings list now sustains ~44 fps instead of ~20.


FlorentRevest:
I love it but isn't that something @moWerk tried and said visibly stuttered when initiating an app movement?

MagneFire:
> I love it but isn't that something @moWerk tried and said visibly stuttered when initiating an app movement?

I'll wait for a reply from him then. But `minnow` is generally really sluggish, these improvements, make it better still the occasional stutter here and there ofc.

moWerk:
> I love it but isn't that something @moWerk tried and said visibly stuttered when initiating an app movement?

Great memory!
And i was thinking the same first. But this actually works. Maybe in my tests i mixed up different tests. Very cool to have this double checked dicretely.

PR #283, opened 2026-07-26
compositor: rasterise the round-screen corner mask once

FlorentRevest:
The corner mask sits on top of the entire system UI, so its cost is paid on every frame the compositor ever draws. It was a full-screen black Rectangle in a layer, sampled back by a ShaderEffect: the layer renders the rectangle into an offscreen texture, then the shader reads that texture back and multiplies it by a mask computed per pixel, every frame.

The texture read carries no information. The source is a constant colour, so texColor is always (0,0,0,1) and the whole thing collapses to an alpha ramp over black that depends on nothing but the size of the screen.

Render it once into a texture and blit that instead. A non-live ShaderEffectSource does exactly this, and it can drive CircleMaskShader unchanged as long as it is handed the constant it used to sample - one opaque black pixel, stretched over the item.

Measured on a Pixel Watch 2 (Adreno 702 pinned to 310 MHz, where devfreq keeps this UI anyway), compositor driven at a steady 60 fps, GPU busy time from KGSL gpubusy, 3 runs x 10 samples with a spread below 0.3 pp:

    overlay alone           0.873 -> 0.283 ms/frame   -68%
    watchface, whole frame  6.363 -> 5.773 ms/frame   -9.3%

The output is unchanged: screenshots taken from this build and from the one before it, back to back, are byte-for-byte identical.

Two alternatives measured and rejected. A QtQuick Shape covers only 21.5% of the screen but costs 0.604 ms/frame with CurveRenderer, more than a full-screen textured quad - on this GPU ALU is expensive and texture sampling is nearly free, so fill-rate reasoning predicts the wrong winner. GeometryRenderer is cheap (0.066) but cannot antialias the arc. A pre-rendered PNG matches the baked texture (0.290) but ties the asset to one screen size.


MagneFire (approved):
Awesome stuff!

PR #285, opened 2026-07-26
Drive Always-on-display wakeup events using a timerfd

FlorentRevest:


FlorentRevest:
This goes hand in hand with https://github.com/AsteroidOS/meta-asteroid/pull/296

lipstick/src/compositor/lipstickcompositor.cpp
```diff
@@ -647,69 +669,59 @@ void LipstickCompositor::setAmbientEnabled(bool enabled)
...
-        return;
+    // Fire at the next wall-clock minute boundary (:00), when the minute digit
+    // changes. Absolute CLOCK_REALTIME time, delivered even through suspend.
+    time_t now = time(nullptr);
+    struct tm tmv;
+    localtime_r(&now, &tmv);
+    tmv.tm_sec = 0;
+    time_t wakeupTime = mktime(&tmv);
+    if (wakeupTime == (time_t)-1) {
+        wakeupTime = now;
```
MagneFire: What would be an instance where the calendar time cannot be represented?
Perhaps it also makes sense to log a warning here?

lipstick/src/compositor/lipstickcompositor.cpp
```diff
@@ -647,69 +669,59 @@ void LipstickCompositor::setAmbientEnabled(bool enabled)
...
-        QDBusReply<uint> res = m_timedDbus->add_event_sync(wakeupEvent);
+
+    // If we are no longer dozing (display came on, or ambient was disabled)
+    // there is nothing to repaint; the arming side disarms on those events,
+    // but guard against a race here too.
+    if (m_currentDisplayState == QMceDisplay::DisplayOn || !ambientEnabled()) {
+        return;
     }
+
+    // The CLOCK_REALTIME_ALARM RTC wakeup keeps the AP awake long enough to
```
MagneFire: Do we actually know how long it stays awake? Iirc in the previous approach MCE held a wakelock.

lipstick/src/compositor/lipstickcompositor.h
```diff
@@ -228,8 +227,13 @@ private slots:
     QWaylandQtWindowManager *m_wm;
     AsteroidGesturesManager *m_gesturesManager;
 
-    Maemo::Timed::Interface *m_timedDbus;
     bool m_ambientModeEnabled;
+
+    // Ambient (AOD) minute tick. An alarm-class timerfd wakes the AP out of
```
MagneFire: A similar comment is also shown in the cpp file. Is there a reason to have this duplicate? Perhaps it's good enough to mention it only once?

lipstick/src/compositor/lipstickcompositor.cpp
```diff
@@ -100,10 +111,17 @@ LipstickCompositor::LipstickCompositor()
...
     connect(QGuiApplication::clipboard(), SIGNAL(dataChanged()), SLOT(clipboardDataChanged()));
 
-    m_timedDbus = new Maemo::Timed::Interface();
-    if( !m_timedDbus->isValid() )
-    {
-      qWarning() << "invalid dbus interface:" << m_timedDbus->lastError();
+    // Alarm-class timer for the ambient (AOD) minute tick. CLOCK_REALTIME_ALARM
+    // wakes the AP out of autosleep, so the watchface still updates off-charger,
+    // and it tracks wall-clock so ticks stay aligned to :00 across NTP/timezone
+    // changes. Requires CAP_WAKE_ALARM, granted to the launcher unit.
```
MagneFire: Have you checked that this is set in the systems service? 

We have the service file also in the meta-smartwatch layer (should probably merge with meta-asteroid someday)

MagneFire:
Forgot to mention but this is great! Keep it up 💪

lipstick/src/compositor/lipstickcompositor.cpp
FlorentRevest: Ah yeah, it's a little confusing because that's done in yet another pull request uh https://github.com/AsteroidOS/meta-asteroid/pull/290/commits

The whole timerfd conversion is to untangle both dsme, iphb and timed so currently it's a bit of a soup but eventually things will be more straightforward to understand hopefully!

lipstick/src/compositor/lipstickcompositor.cpp
FlorentRevest: Was over careful slop, dropped.

lipstick/src/compositor/lipstickcompositor.cpp
FlorentRevest: Hopefully the new explanation is clearer. The kernel's alarmtimer mechanism seems to hold a wakeup source when an alarm-class timer fires, until the fd is serviced. As far as I could test on my aurora, the frame rendering seems to happen within that time frame so. Maybe there are some oddities left, we should keep an eye on it but that seems to work for me. At least better than the timed case which was broken on aurora for some reason

lipstick/src/compositor/lipstickcompositor.h
FlorentRevest: Done

lipstick/src/compositor/lipstickcompositor.cpp
FlorentRevest: Mh well, I've been arguing with my slopbot and it's actually unsure at this point. My personal take is: it seems to work without a wake lock at the moment and introducing so I'm tempted to go ahead with this and see what happens. If it turns out to be an issue, we can consider introducing a wakelock but that will require CAP_BLOCK_SUSPEND too and maybe exposing some sysfs files too so I'm not too keen right unless that's proven necessary.

dodoradio: If this ends up needing a wakelock, can we please avoid using the wakelock kernel API directly, and instead talk to MCE over DBUS? Obviously this would be another dependency on MCE, but we have plenty of those.
The kernel wakelock API is just not very good (it relies on writing a string into a `lock` and `unlock` files). It feels a lot safer if we just have one daemon that does all the wakelocking.

dodoradio:
I feel a bit mixed on this. As I mentioned in a comment above, the timerfd and wakelock APIs are a bit austere, and they're not something that applications would want to use. 
The current model, where timed does the wakeups, is not ideal. However, overall, I think there is value in having something like timed which abstracts wakeups. 
I'm not sure it's good that this functionality is being brought directly into asteroid-launcher, as it would imply that every other application that wants to do anything in the background should also be talking to timerfds directly. 

Of course, the AOD usecase is a little bit special, as it needs to happen exactly once a minute and it happens more frequently than most other background tasks would. So maybe this solution is appropriate here, but some thought does need to be given to background tasks would work elsewhere.

FlorentRevest:
> I feel a bit mixed on this. As I mentioned in a comment above, the timerfd and wakelock APIs are a bit austere, and they're not something that applications would want to use.
> The current model, where timed does the wakeups, is not ideal. However, overall, I think there is value in having something like timed which abstracts wakeups.
> I'm not sure it's good that this functionality is being brought directly into asteroid-launcher, as it would imply that every other application that wants to do anything in the background should also be talking to timerfds directly.

If your concern is API complexity, we could trivially have a C++ library that wraps timerfd's low level file descriptor APIs and what not with a higher level API (like QTimer). The part that matters to me is that we no longer need complex IPCs and a privileged daemon to do such simple things. I don't care what abstraction level we use to communicate with timerfd, what matters to me is that architecturally it's a lot simpler to have each process be able to directly ask the kernel for wake up events using a mainline API rather than have this crazy iphb custom socket protocol communicate with dsme which itself has crazy abstractions for legacy wakeup mechanisms that none of our watches have.

MagneFire (approved):

MagneFire:
@dodoradio Are you ok with the latest comments?

dodoradio:
I think AOD can be a good reason to talk to the kernel directly, since we can almost see it as a hardware feature. So yeah, I think this is good to merge!

Regarding other usecases however, a response to kido's comment:
> If your concern is API complexity, we could trivially have a C++ library that wraps timerfd's low level file descriptor APIs and what not with a higher level API (like QTimer). The part that matters to me is that we no longer need complex IPCs and a privileged daemon to do such simple things. I don't care what abstraction level we use to communicate with timerfd, what matters to me is that architecturally it's a lot simpler to have each process be able to directly ask the kernel for wake up events using a mainline API rather than have this crazy iphb custom socket protocol communicate with dsme which itself has crazy abstractions for legacy wakeup mechanisms that none of our watches have.

I think some level of abstraction is actually desirable. Not just an API wrapper, but something that abstracts some of the design decisions from application developers. 
I think part of the concept of timed, a daemon that schedules wakeups based on requests from applications over DBUS, is desirable. broadly because:
- Only one daemon needs to talk to the kernel
- Talking to it over DBUS means that applications could continue to work within, eg., flatpak
- That daemon can make smart decisions, eg. about grouping wakeups. The API could be a bit more abstract than 'when' or 'how often'. Instead, wakeups could be scheduled as 'frequent', 'regular' or 'occasional', and then the daemon could schedule actual wakeups based on user activity and system resources, such as battery or network state. Wakeups could also depend on the availability of things like the network state (eg. not attempting to sync online data when the system is offline).

Obviously timed doesn't do all of this, and it has accumulated issues in both design and implementation, like the dependency on DSME, or the very rigid and tangled alarmclock system. But I don't think it's worth completely throwing out the idea of wakeup middleware, as it seems kinda necessary for more complex background usecases, and it's something that seems to be missing on other mobile linuxes. (Apparently systemd.timer hits some of these usecases, but I've not seen any well-developed implementations of power-aware background sync with it, so I'm not sure)

PR #291, opened 2026-08-03
compositor: host the virtual keyboard and fit it to round screens

FlorentRevest:
Qt6 removed client-side virtual keyboard support under Wayland, so add the keyboard to the compositor and let clients reach it over the Qt text-input protocol:

- Register QWaylandQtTextInputMethodManager in the LipstickCompositor constructor to advertise qt_text_input_method_v1. It can't be nested in the Compositor QML item: that subclass derives from QWaylandQuickCompositor directly and lacks the "data" default property the quick extensions rely on, so it is created in C++ alongside the existing XdgShell / QtWindowManager / gestures extensions.
- Add a bottom-sliding InputPanel to the compositor scene. On round screens it is inset to 86% width and centered so the top-row Q and P keys stay inside the circle instead of being clipped by the corner mask.

Considerably more work would be required to style this keyboard better for watch screens but this is at least an incremental step in the right direction.


FlorentRevest:
This goes hand in hand with https://github.com/AsteroidOS/meta-asteroid/pull/310

PR #292, opened 2026-08-04
Fix live reload reverting a just-activated face

moWerk:
f8e47ba reloads the homescreen face after a store install by bouncing the dconf-backed watchFaceSource (save it, clear it, write it back). 
But because that key is dconf-backed, the write-back races the settings app, which writes the same key to activate the just-installed face in another process. When the ~/.fonts install event reaches the launcher before the activation write does, onReloadNeeded writes the previous face back over the activation: the selection reverts and the compositor never loads the new one. Faces that bundle a font are affected, and whether an install sticks is pure timing.

This routes a signal from the desktop root into the watchface Loader and reinstantiates it to re-read the cache-cleared source, so the live reload keeps working without ever writing watchFaceSource, and a concurrent activation is never clobbered.

Review ask: If the comments feel too elaborate or not timeless enough, i am happy to rework.

Tested on beluga and catfish: community faces that intermittently reverted during development/testing of the incoming preview-auto-generator PR (e.g. navigator, halloween, duppy-vintage) now stay activated, and selection is noticeably snappier.

### DIFF
--- a/src/qml/MainScreen.qml
+++ b/src/qml/MainScreen.qml
@@ -48,6 +48,10 @@ Item {
     visible: Lipstick.compositor.homeActive || aboutToOpen || aboutToClose || aboutToMinimize
     enabled: visible
 
+    // Requests the active watchface Loader to reload from disk. A signal, so
+    // the handler can live in the Loader's own scope.
+    signal watchfaceReloadRequested()
+
     AppLauncherBackground { id: alb }
 
     property var defaultCenterColor: alb.centerColor("/usr/share/asteroid-launcher/default-colors.desktop")
@@ -365,9 +369,7 @@ Item {
         // dropped the component cache, so re-trigger the loader to re-read the
         // directory and pick the new face up live.
         function onReloadNeeded() {
-            var watchFaceSourceBackup = watchFaceSource.value
-            watchFaceSource.value = ""
-            watchFaceSource.value = watchFaceSourceBackup
+            desktop.watchfaceReloadRequested()
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/qml/MainScreen.qml
```diff
@@ -365,9 +371,7 @@ Item {
         // dropped the component cache, so re-trigger the loader to re-read the
         // directory and pick the new face up live.
         function onReloadNeeded() {
-            var watchFaceSourceBackup = watchFaceSource.value
-            watchFaceSource.value = ""
-            watchFaceSource.value = watchFaceSourceBackup
+            desktop.watchfaceReloadRequested()
```
FlorentRevest, line 372: Why not do the active = true and active = false directly here?

src/qml/MainScreen.qml
```diff
@@ -48,6 +48,12 @@ Item {
     visible: Lipstick.compositor.homeActive || aboutToOpen || aboutToClose || aboutToMinimize
     enabled: visible
 
+    // Emitted when a store install lands a new face/font and the loader must
+    // re-read it. Handled inside the watchface Loader (see centerPanel) by
```
FlorentRevest: Don't refer to specific uses because they might change over time and this comment will become stale

src/qml/MainScreen.qml
```diff
@@ -393,6 +397,19 @@ Item {
                 Behavior on opacity { NumberAnimation { duration: 200; easing.type: Easing.InCirc} }
                 anchors.fill: parent
                 source: watchFaceSource.value
+
+                // A store install cleared the component cache and asked for a
+                // live reload. Reinstantiate to re-read the source from disk,
+                // without touching watchFaceSource: writing that dconf key here
+                // races the settings app's activation write and can clobber a
```
FlorentRevest: Only mention this in one of the two comments

src/qml/MainScreen.qml
FlorentRevest: Ah nevermind now I read more code and see it's in a component... OK

src/qml/MainScreen.qml
moWerk: Much appreciated! fixed to only one mention.

src/qml/MainScreen.qml
moWerk: Yeah, thats me seeing this to much in light of the topic this was found in. Noted.

PR #295, opened 2026-08-07
Replace the stock preview PNGs with transparent webp masters

moWerk:
Replaces every stock watchface's five per-size opaque preview PNGs with a single transparent webp master, mirroring the community migration in AsteroidOS/unofficial-watchfaces#262. The store scales the master to the device and composites it over the live wallpaper, and the launcher's preview renderer (#293) regenerates a device-sized grab whenever a face is activated, so no per-resolution set needs shipping. 001-words gains its first preview of any kind. Until now its store tile could only render the face live.

Each master is a transparent 192 pixel square, which covers the largest display in the fleet at the proportion a store tile draws. Faces whose knockout designs reveal the wallpaper through shaped holes in a dark layer, 016-masked-spartan among them, keep that transparency intact: the dark layer stays partially opaque and the holes are fully clear, so the tile shows the wallpaper through the face exactly as the homescreen does.

85 PNGs removed, 18 masters added, one commit. Part of the watchface store set (AsteroidOS/unofficial-watchfaces#44).

This must land together with AsteroidOS/asteroid-settings#145, which reads these masters. Merging this alone removes the per-size PNGs that the current store resolves, leaving stock tiles to render live.

This was executed by an LLM under my direction. I have fully understood and take authorship of the changes.


MagneFire (approved):

MagneFire:
Guessing this work depends on https://github.com/AsteroidOS/asteroid-launcher/pull/293?

PR #299, opened 2026-09-13
Rotate watch in nightstand mode

MagneFire:
Some charging cradles stand the watch on its side and turn it into a bedside clock.

Rotate the UI with the watch while nightstand mode is active.
Ambient mode pauses the sensor but keeps the orientation, so the clock stays rotated while dimmed.


MagneFire:
This implements https://github.com/AsteroidOS/asteroid/issues/151

MagneFire:
This also depends on https://github.com/AsteroidOS/meta-asteroid/pull/329 for correct rotation if the watch was rotated before switching to nightstand mode.

